Issue #17, opened 2015-09-17
Compatibility between Antispam Bee and WPML / Mail from WPML


staude:
I received this Mail from WPML 

---

Dear Frank,

My name is Helena from WPML.  We have received a request in our support forum to check on the compatibility of Antispam Bee with WPML  which you may find here:
 https://wpml.org/de/forums/topic/kommentare-schreiben-nicht-moeglich-bei-fremdsprache/

Description by the assigned supporter:

 When WPML and Antispam Bee are activated the comments in 2nd language are not working :

 https://d2salfytceyqoe.cloudfront.net/wp-content/uploads/2015/09/704843-Bildschirmfoto_2015_09_14_um_20.48.24.png
 https://d2salfytceyqoe.cloudfront.net/wp-content/uploads/2015/09/704843-Bildschirmfoto_2015_09_14_um_20.48.33.png

In order to get  Antispam Bee fully compatible with WPML,   we would like to invite you  to join our  GoGlobal  program which is specially designed for developers.

http://wpml.org/documentation/theme-compatibility/go-global-program/

Simply complete the details and then send directly to me your latest version by mail.    We will set up a test site with your theme/plugin installed where we can work on the issues together, test it and give you a detailed report of the results.

After the process is complete, we will announce it to our users and add Antispam Bee  to our compatibility page.

You can see the successful list here

http://wpml.org/documentation/theme-compatibility/

http://wpml.org/documentation/plugins-compatibility/

Looking forward to hearing from you,

Helena Petridou
WPML administrative assistant
OnTheGoSystems
www.onthegosystems.com
The makers of WPML, Toolset and ICanLocalize

swissspidy:
So Antispam Bee doesn't work correctly on multilingual sites, no matter what plugin is used? Or is it just restricted to WPML?

The "Allow comments only in certain language" setting isn't available at the moment, so that can't be the culprit.

staude:
I have sent Helena the URL to this ticket and hope that expresses itself. even I can not say anything, except that I got the mail.

swissspidy:
Closing this as per our recent discussion on Slack.

nadalsol:
Sorry to bother you @swissspidy but has this issue been solved?

I'm having this exact problem on my site:
https://wordpress.org/support/topic/problem-with-multi-language-site/

WordPress v 4.6.1
Anti Spam Bee v 2.6.9

Thanks for your time

swissspidy:
[...]

PR #24, opened 2015-10-18
bbPress Antispam integration

danielhuesken:
I have iproved the secret key generation so that it is not predictable and change every month automaticly.

I have spent it to new moethods for reuse on later bbPress Antispam integration.

Please test again bevore pull.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1788,15 +1788,18 @@ class Antispam_Bee {
 		}
 
 		/* bbPress DB */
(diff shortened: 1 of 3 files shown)


danielhuesken:
Can be that it fixes #17 too.

danielhuesken:
Now, the intrgration of bbPress Antisam is ready for test.

swissspidy:
IIRC there was some discussion on Slack about this being more suitable for a separate plugin?

krafit:
We've been carrying this PR around for some time now. I don't see bbPress Support in ASB until v3.0, so I'll close this PR for now. Would love to see this as an Add-on in the future!

PR #40, opened 2016-09-09
cleaned changelog +support and contribute added

timse201:
think release time investment could be cleaned, too?

will create a CHANGELOG.md later this day

PR #41, opened 2016-09-09
Create CHANGELOG.md

timse201:
some german/english translation is missing
will add this today or tomorrow ;)


swissspidy:
Would you mind adding this as part of #40? Removing the changelog from one file and putting it in the other should clearly be 1 PR.

Issue #15, opened 2015-09-17
Remove translation files when possible


swissspidy:
We can safely remove any translation files from the plugin here that reached 100% on WordPress.org. 
So far only German is fully translated: https://translate.wordpress.org/projects/wp-plugins/antispam-bee/dev

This is a tracking ticket we can keep open a bit to slowly remove translation files and of course bring people to translate on WordPress.org.

PR #42, opened 2016-09-09
remove slovenian and french translation

timse201:
after pushing the remaining 4 translations we can delete these files

[WordPress translate](https://translate.wordpress.org/projects/wp-plugins/antispam-bee/stable/sl/default?filters%5Btranslated%5D=yes&filters%5Bstatus%5D=waiting)


timse201:
a French translation is already available
all strings in the deleted files are already in the files on WordPress Translate

swissspidy:
Well, there's a French language pack, but not a Slovenian one because it's not complete yet. This means we can only remove the fr_FR files at the moment.

See https://translate.wordpress.org/projects/wp-plugins/antispam-bee/language-packs

swissspidy:
By the way, it's easier for us to handle issues if you mention the issue numbers when opening a PR.  That way the issue gets updated automatically when merging a PR.

In this case, this PR would be about #15.

timse201:
true the first post

after pushing the remaining 4 translations we can delete these files,too

swissspidy:
Can you undo 1e360e823b1ed5cdd6787a88477f572534085146 and 291531a93e2afb0d3bd802bbee6e1dfcffea88d8 so I can merge this PR? Or just a open a new PR to only delete the FR files, whatever works best for you.

timse201:
did both into a new one

PR #43, opened 2016-09-12
delete slovenian translation

timse201:
after pushing the remaining 4 translations we can delete these files

[WordPress translate](https://translate.wordpress.org/projects/wp-plugins/antispam-bee/stable/sl/default?filters%5Btranslated%5D=yes&filters%5Bstatus%5D=waiting)
#15


Zodiac1978:
Unfortunately the translated strings are not validated yet, so the translation remains at 92% and the language pack wouldn't be delivered (this just happens with 95%+), so we should wait for the validation of the last four strings.

I try to ping some slovenian GTE.

Zodiac1978:
@timse201 Why did you closed this?

timse201:
we can open it again when its finished
and 1 is now fuzzy so the slovenian gte seems to have checked it

https://translate.wordpress.org/projects/wp-plugins/antispam-bee/dev/sl/default?filters%5Bstatus%5D=fuzzy

PR #44, opened 2016-09-12
delete french translation

timse201:
a French translation is already available
all strings in the deleted files are already in the files on WordPress Translate
#15


swissspidy:
French is at 96% on translate.wordpress.org, so a language pack is already available.

Issue #18, opened 2015-09-22
Improve default Regex


swissspidy:
We could add the patterns from https://github.com/sdellenb/twentyfifteen-rubenesque-child/blob/master/functions.php#L14-L45

glueckpress:
+1

PR #48, opened 2017-01-06
More default Regexes

websupporter:
Solves https://github.com/pluginkollektiv/antispam-bee/issues/18

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1527,9 +1527,6 @@ class Antispam_Bee {
 				'body'	=> '^dating|sex|lotto|pharmacy$',
 				'email'	=> '@mail\.ru|@yandex\.$',
 			),
-			array(
-				'body' => '^(?=.{0,2}$).*'
-			),
 		);
 
 		/* Spammy author */
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1527,6 +1527,9 @@ class Antispam_Bee {
 				'body'	=> '^dating|sex|lotto|pharmacy$',
 				'email'	=> '@mail\.ru|@yandex\.$',
 			),
+			array(
+				'body' => '^(?=.{0,2}$).*'
+			),
 		);
 
 		/* Spammy author */
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1515,7 +1515,7 @@ class Antispam_Bee {
 				'body'	=> '\<\!.+?mfunc.+?\>',
 			),
 			array(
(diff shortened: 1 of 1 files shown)


Zodiac1978:
I would like to see some documentation for this. (e.g. non-english words)

There is a way to provide non detected spam comments:
https://github.com/pluginkollektiv/antispam-bee/wiki/de-H%C3%A4ufige-Fragen#nicht-erkannten-spam-einreichen
https://docs.google.com/forms/d/e/1FAIpQLSeQlKVZZYsF1qkKz7U78B2wy_6s6I7aNSdQc-DGpjeqWx70-A/viewform?c=0&w=1

Here are the results:
https://docs.google.com/spreadsheets/d/1IJWa8nBlGasJhJbA3fvGnh-QPt4EhGTA9w5eZByGYs0/edit#gid=260223063

Would be great to add even more patterns.

schlessera:
Matching substring `dress` as author seems to invite lots of false positives, as it can be part of legitimate names, like `Andressa`.

Also, why is `prada` included in Japanese, but not in English?

Zodiac1978:
See #18 - this is just adopted from this child theme. If you want the answer to that question you have to ask the original developer ;)

I think we should remove "dress" then and include prada.

Furthermore we can add words like cialis, viagra, etc.

websupporter:
Yes, it was just quickly done. Will add/remove as suggested later in the evening.

sdellenb:
The RegExes are based on the spam I got a few years back. For some reason, I never got undetected spam that had "Prada". Maybe some other test kicked in first.

In the mean time, I've also added [viagra|cialis](https://github.com/sdellenb/twentyfifteen-rubenesque-child/blob/master/functions.php#L46), because such spam comments started to appear. 👍

websupporter:
Thanks @sdellenb,
I have added those now to the body param too. Forgot about this possibility :smile:

websupporter:
I did add the pattern found @Zodiac1978 blog: http://torstenlandsiedel.de/2017/01/08/custom-pattern-fuer-antispam-bee/

This marks comments not longer than two chars as spam.

Zodiac1978:
We could check if the IP and Hostnames are valid:
http://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address

And we could check if there is an IP or email address instead of a regular Website given. Or is that too much?

websupporter:
Basically the last commit removes the `2 char`-commit to enable Comments like "Ja", which are completly legit. Following the remarks here: https://github.com/Zodiac1978/custom-patterns-asb

PR #50, opened 2017-01-08
New Links, Link correction, PayPal -> Donate

timse201:


inc/gui.class.php
```diff
@@ -160,74 +160,74 @@ public static function options_page() { ?>
...
 								</label>
 							</li>
 
 							<li>
 								<input type="checkbox" name="ab_regexp_check" id="ab_regexp_check" value="1" <?php checked($options['regexp_check'], 1) ?> />
 								<label for="ab_regexp_check">
-									<?php esc_html_e('Use regular expressions', 'antispam-bee') ?>
-									<span><?php _e('Predefined and custom patterns by <a href="https://gist.github.com/4242142" target="_blank">plugin hook</a>', 'antispam-bee') ?></span>
+									<?php esc_html_e( 'Use regular expressions', 'antispam-bee' ); ?>
+									<span><?php esc_html_e( 'Predefined and custom patterns by <a href="https://gist.github.com/4242142" target="_blank" rel="noopener noreferrer">plugin hook</a>', 'antispam-bee' ); ?></span>
```
Zodiac1978: This link is dead. We should remove it, because we have no replacement at the moment.

Zodiac1978: And as the translation contains HTML I think we can't escape it, because it would show the escaped link instead of link then.

inc/gui.class.php
```diff
@@ -160,74 +160,74 @@ public static function options_page() { ?>
...
 								</label>
 							</li>
 
 							<li>
 								<input type="checkbox" name="ab_dnsbl_check" id="ab_dnsbl_check" value="1" <?php checked($options['dnsbl_check'], 1) ?> />
 								<label for="ab_dnsbl_check">
-									<?php esc_html_e('Use a public antispam database', 'antispam-bee') ?>
-									<span><?php _e('Matching the ip address with <a href="http://www.stopforumspam.com/" target="_blank">Stopforumspam</a>', 'antispam-bee') ?></span>
+									<?php esc_html_e( 'Use a public antispam database', 'antispam-bee' ); ?>
+									<span><?php esc_html_e( 'Matching the ip address with <a href="http://www.stopforumspam.com/" target="_blank" rel="noopener noreferrer">Stopforumspam</a>', 'antispam-bee' ); ?></span>
```
Zodiac1978: See above. I think, we can''t escape the translation as this would result in showing the escaped link and not showing a clickable link.

Issue #25, opened 2016-01-12
Harmonize admin UI


glueckpress:
I’d like to implement a couple of enhancements on ASB’s settings page, mostly to support fluid/responsive layout, varying font-sizes etc. As of today, the UI looks neat, but breaks when e.g. feature descriptions exceed a certain line number. 

Will make a list of specific enhancements soon. 
Also open for feedback. 
(Note: We’re talking about improving existing UI, not new features or anything. :wink: )

Zodiac1978:
@timse201 @glueckpress Can we close this issue or should it stay open for more improvements?

I would like to release 2.7 soon.

timse201:
the only thing i would like to do is to include the privacy notes into the descriptions
will do a pull request soon
but please note #49

timse201:
after all this set i think it can be closed
the ui does not break translations anymore
the ui looks the same on mobile and desktop devices
and there are some advantages now like input number

PR #51, opened 2017-01-09
nowrap to max-width

timse201:
Input remains in position but the text can make a line break

some translations could be longer than the space that has been given. The description can now be longer and the privacy notice could be in the description.

#25


websupporter:
Played a bit with it. works fine for me.

PR #53, opened 2017-01-15
more tags; 4.7

timse201:


Zodiac1978:
I'm fine with the changes, but we should know that
1. we have a recommended maximum of 12 tags in the directory:
https://github.com/Ipstenu/wporg-plugin-guidelines/blob/master/guideline-13.md
and 2. it could be revised in the future. More info here: 
https://make.wordpress.org/plugins/2016/02/25/re-thinking-tags-in-the-plugin-directory/

PR #56, opened 2017-01-16
#comment & unique secret

websupporter:


schlessera:
If I understand the code correctly, you'll end up with two textareas having the same ID, which is invalid HTML. Was that intended?

Also, it would be great to have a public static method to fetch the current secret, so that other plugins can use that to integrate with ASB.

websupporter:
Yes, I thought this was the agreement in the end. But I am not very happy with it too.

Zodiac1978:
At the moment you can build the secret yourself if you know the code. It is the first 5 chracters of md5 hash of the public info "Site-URL". We could try to change the `get_bloginfo('url')` to `NONCE_SALT` so we have a real *secret* information which is not easy to guess from the outside. Maybe with a different prefix length (5 -> 6). Just to change something.

Did the secret has to stay the same for every comment or could we alternate this for each page view?

Zodiac1978:
Additionally I think we couldn't know if `NONCE_SALT` exists. We should test that and use a fallback if it is empty.

schlessera:
We need a way for other plugins to fetch this information without needing to know implementation details (so we can change the implementation as needed). Other plugins that hook into the commenting system need the information to properly include support for ASB.

websupporter:
What do you think, can we use `ABSPATH` instead of `NONCE_SALT`?

schlessera:
`ABSPATH` is somewhat predictable. Maybe something like this just to have a fallback value:
```php
$salt = defined( 'NONCE_SALT' ) ? NONCE_SALT : md5( ABSPATH );
```

schlessera:
Sorry, my bad, I shouldn't have included the `md5()` in the example... :)

websupporter:
No worries :)

websupporter:
The latest commits add a quite unique ID to the "good" comment field. A JS swapps the IDs from the good to the bad comment field.

props @2ndkauboy

Zodiac1978:
What about [wp_hash](https://developer.wordpress.org/reference/functions/wp_hash/)? Coudl'nt we use that? Or [wp_salt](https://developer.wordpress.org/reference/functions/wp_salt/).

If there is a self-learning system on the other side, maybe we should make this dynamic. We could add the date to the string. Just one problem: You open the page on day 1 and send the comment on day 2. If this happens we should ask the user to re-send the comment, maybe after clicking a confirmation checkbox. This could make the CSS hack working again.

Are there any other possibilities to hide an element besides `display: none;` which do not "showing" the element screenreaders? Maybe in combination with `ARIA hidden `? It seems that the spammer software recognizes honeypots which are not displayed.

websupporter:
Hust a quick thought:

> We could add the date to the string

The other problem would be Caching.

websupporter:
Instead of `display:none` the latest commit moves it out of the window. I am not quite sure, if this would play nicely in every layout environment. But it is ridiciously far left. I think we could give it a try `aria-hidden` is set.

schlessera:
How about putting the honeypot field directly behind the created field?

websupporter:
I think it needs an `!important`, but I can't see the benefit of moving it under the other `textarea`.

schlessera:
I was thinking about how a tool like XRumer would try to intelligently detect a honeypot<=>real field presence. My first guess would be "hidden or moved out of view"...

websupporter:
My idea would be to merge it now. Its an improvement, but should be tested with the other changes together again. Some CSS can be changed later easily. What do you think @schlessera

schlessera:
@websupporter Can you confirm that the above implementation still works as expected regarding the honeypot field?

websupporter:
@schlessera I was writing and testing it and I am confident. But since it's quite a change in the core of the system, I would be happy, if we merge it early in the testing phase. More eyes are helpful here.

Issue #57, opened 2017-01-16
JS Nonce


websupporter:
We could introduce an optionally JS feature.

We create a hidden field, which will get populated with a WP Nonce by Javascript. If the input fields value does not match: SPAM.

swissspidy:
How safe is this? When you have a public route to get a nonce, what stops an attacker from fetching a nonce and populating a field with that nonce?

websupporter:
Hi @swissspidy,
basically nothing, but they have to do it. So its another step, which needs to be taken. In the current PR, you need to know the public route and a nonce (`wp_rest`), so you would have to do some action, to get the correct nonce. But: This introduces a problem with cached pages right now and I think this PR needs some more thought.

Right now, it is not very high on my agenda. But I didn't want to close it, since this might be a feature, worth considering.

websupporter:
wont fix for now.

PR #58, opened 2017-01-16
Js Nonce

websupporter:
As suggested in https://github.com/pluginkollektiv/antispam-bee/issues/57 I have created a small optional feature.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -72,30 +72,32 @@ class Antispam_Bee {
   			)
   		);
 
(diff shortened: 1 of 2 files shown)


websupporter:
- Needs correct & German (?) PHP Documentation.
- Since we do send the `wp_rest` Nonce, I think we do not need to send a second nonce to the Ajax for the trap.
- Register Ajax-Handlers only, when the option is activated.
- There is still working test code in it.

websupporter:
wont fix for now.

Issue #55, opened 2017-01-16
Test suite (unit / integration / acceptance)


schlessera:
The plugin needs a test suite to let us more confidently make changes to the existing code. This avoids regressions, and could for example be used to run through the list of known spam so we can optimize against this.

I'll create a pull request with the beginnings of such a test suite.

Some considerations:

- To pull in the testing framework(s), Composer should be used, that's why I will add a `composer.json` file. I will only add stuff for the development environment for now, and this can be completely stripped during the release packaging.

- The code is not built in a modular way, so there will be a lot of fiddling to even get to the point where a test can be run. Over time, we should strive to modernize the code so that testing gets easier as well.

schlessera:
This has been merged now.

To install all the tools necessary for running the test suite, you need to let Composer pull it ins development dependencies:
```composer install```
To run the entire test suite:
```vendor/bin/phpunit```
To run only one of the suites (unit, integration or acceptance), add the --testsuite=<suite> switch:
```vendor/bin/phpunit --testsuite=unit```

PR #59, opened 2017-01-17
Add early implementation of a complete test suite. See #55

schlessera:
To install all the tools necessary for running the test suite, you need to let Composer pull it ins development dependencies:
```
composer install
```

To run the entire test suite:
```
vendor/bin/phpunit
```

To run only one of the suites (unit, integration or acceptance), add the `--testsuite=<suite>` switch:
```
vendor/bin/phpunit --testsuite=unit
```

Note: The tests will only pass once #48 has been merged.


schlessera:
Just for the record: This does not do much yet, the initial effort was to even get to the point where a spam reason can be detected and tested.

PR #85, opened 2017-02-02
Swap `parse_url()` for `wp_parse_url()`

### LINKED ISSUE #77: Avoid `parse_url`
All the instances of `parse_url()` should be replaced with `wp_parse_url()`.

`parse_url()` has known inconsistencies across PHP versions.

schlessera:
Fixes #77.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1087,7 +1087,7 @@ class Antispam_Bee {
 
 		/* Request params */
 		$request_uri = self::get_key($_SERVER, 'REQUEST_URI');
-		$request_path = parse_url($request_uri, PHP_URL_PATH);
+		$request_path = wp_parse_url($request_uri, PHP_URL_PATH);
 
 		/* Request check */
 		if ( strpos($request_path, 'wp-comments-post.php') === false ) {
@@ -1135,7 +1135,7 @@ class Antispam_Bee {
 
 		/* Request params */
 		$request_uri = self::get_key($_SERVER, 'REQUEST_URI');
-		$request_path = parse_url($request_uri, PHP_URL_PATH);
+		$request_path = wp_parse_url($request_uri, PHP_URL_PATH);
 
 		/* Empty path? */
 		if ( empty($request_path) ) {
@@ -1300,7 +1300,7 @@ class Antispam_Bee {
 		}
 
 		/* IP != Server */
-		if ( $options['advanced_check'] && self::_is_fake_ip($ip, parse_url($url, PHP_URL_HOST)) ) {
+		if ( $options['advanced_check'] && self::_is_fake_ip($ip, wp_parse_url($url, PHP_URL_HOST)) ) {
 			return array(
 				'reason' => 'server'
 			);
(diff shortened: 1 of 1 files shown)


websupporter:
I think it makes sense to use it also here
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L1303
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L1418

chesio:
According to documentation, [wp_parse_url](https://developer.wordpress.org/reference/functions/wp_parse_url/) is available from WordPress 4.4, so required WP version ([3.8](https://github.com/pluginkollektiv/antispam-bee/blob/master/readme.txt#L5) in the moment) should be bumped before this is merged .

timse201:
True
I think we should think about some stricter min versions why should be support outdated software (End Of Life) with security vulnerabilitys
shouldnt we say [min php 5.5](http://php.net/supported-versions.php)
and [WP 4.4](https://de.wikipedia.org/wiki/WordPress#Versionen)

so we can say we tested it 1 version more and give a little hint that others could be also work but are untested like bueltge already requested [here and did for publish confirm](https://github.com/pluginkollektiv/publish-confirm/issues/15)

schlessera:
@websupporter These are already included in the initial PR.

@chesio Thanks for noticing this, I have to admit I didn't check the introduction of that function.

As we don't have had immediate issues with this, I suggest we leave this patch out for the imminent release, then, to not confuse people with unnecessary update requirements.

websupporter:
@chesio good catch.

krafit:
@timse201 since WordPress introduced autoupdates back in 3.7 all major versions since receive security updates. So "security vulnerabilitys" dose not actually apply to previous WP versions 🙃 

Bumping supported PHP Versions on the other hand might be a step worth taking. But without actual data it's a thought decision. We should spend some more time and thought on this.

websupporter:
If we finally decide to go to another WP- or PHP-version, we should die gracefully I think. Nobody wants users complaining about after update white screen and stuff.

swissspidy:
Ps. Always use something like "Fixes #xy" in the PR description. That way an issue gets referenced &  automatically closed when the PR gets merged.

Regarding the version requirements: this is out of scope for this PR. Especially since PHP version requirement isn't affected by this PR. Feel free to open an issue about it where we can have a discussion specifically about.

The WordPress minimum requirement would need to be bumped carefully, with over 200'000 installs there are likely many users on older versions. Theoretically, this would be something for a major release.

However, can't we just use `function_exists( 'wp_parse_url' ) ? wp_parse_url( $url ) : parse_url( $url )` everywhere?

schlessera:
> Ps. Always use something like "Fixes #xy" in the PR description. That way an issue gets referenced & automatically closed when the PR gets merged.

I already had that in the title.

> However, can't we just use function_exists( 'wp_parse_url' ) ? wp_parse_url( $url ) : parse_url( $url ) everywhere?

Goog idea with the conditional swap, I'll update the PR (with a child method to keep it DRY).

websupporter:
Actually, by accident we switched to min version 4.5 seemingly without bigger problems, so we could proceed here.
I removed the blocked label and did put the PR into the milestone 2.8.2 for this reasons.

websupporter:
started this from scratch in #239 to get it easier done, if that's ok.
will close this issue in favour of #239

PR #86, opened 2017-02-02
Allow commas and semicolons in country code list. Resolves #80

### LINKED ISSUE #80: Make country black/whitelist splitting more robust
Currently, the country whitelist and blacklist options from the country check need the list to be separated by spaces. An option like `RU,CH` or `CH;RU` will fail to work.

The code to split the string into separate codes should be made more robust to allow for the most common ways of putting a collection into a string.

schlessera:


timse201:
Please keep placeholders in mind ;)

Issue #99, opened 2017-02-03
<label for> does not point to secret.


websupporter:
* correct the `for`-attribute

enhancement
* move the element on the bottom
* adjust the JS `nextAll()`-method to exclude the element.

PR #100, opened 2017-02-03
correct for-attribute, move secret-input down

websupporter:
Solves #99


websupporter:
get rid of the input-field. its useless. add a filter instead.

websupporter:
put two filter in the `get_secret`-methods instead. more useful.

antispam_bee.php
```diff
@@ -2621,7 +2626,19 @@ public static function get_secret_for_post( $post_id ) {
...
+		$secret = substr( sha1( md5( 'comment-id' . self::$_salt . get_the_title( (int) $post_id ) ) ), 0, 10 );
+
+		/**
+		 * Filters the secret for a post, which is used in the textarea id attribute.
+		 *
+		 * @param string $secret The secret.
+		 * @param int    $post_id The post ID.
+		 */
+		return apply_filters(
+			'get_secret_id_for_post',
```
schlessera: Prefix is missing here.

antispam_bee.php
```diff
@@ -2609,7 +2601,20 @@ private static function _update_daily_stats()
...
+		$secret = substr( sha1( md5( self::$_salt . get_the_title( (int) $post_id ) ) ), 0, 10 );
+
+		/**
+		 * Filters the secret for a post, which is used in the textarea name attribute.
+		 *
+		 * @param string $secret The secret.
+		 * @param int    $post_id The post ID.
+		 */
+		return apply_filters(
+			'ab_get_secret_for_post',
```
schlessera: Maybe add `name` in the filter?

websupporter:
refactored `get_secret_for_post()` => `get_secret_name_for_post()`
added an `esc_attr()`
filter naming fixed

antispam_bee.php
```diff
@@ -1101,13 +1101,13 @@ public static function precheck_incoming_request()
...
-		$hidden_field = self::get_key($_POST, 'comment');
-		$plugin_field = self::get_key( $_POST, self::get_secret_for_post( $post_id ) );
+		$hidden_field = self::get_key( $_POST, 'comment' );
+		$plugin_field = self::get_key( $_POST, self::get_secret_name_for_post( $post_id ) );
 
 		// Hidden field check
 		if ( empty($hidden_field) && ! empty($plugin_field) ) {
 			$_POST['comment'] = $plugin_field;
-			unset( $_POST[ self::get_secret_for_post( $post_id ) ] );
+			unset( $_POST[ self::get_secret_name_for_post( $post_id ) ] );
```
schlessera: Shouldn't this be `unset( $plugin_field )`?

websupporter: We could unset this too. But I think the general idea here is more get rid of the extra `$_POST`-field after the data has been transfered to `$_POST['comment']`

schlessera: Ah, nevermind, I misread the code.

Issue #107, opened 2017-02-04
Update readme.txt for 2.7.0


websupporter:
The readme.txt needs to be adjusted.

schlessera:
Resolved with #109

PR #109, opened 2017-02-04
update readme, update version

websupporter:
solve #107


timse201:
please add some parts here https://github.com/pluginkollektiv/antispam-bee/blob/master/CHANGELOG.md
too

and it should be the same text in both

websupporter:
Lets have a look,if we need to adjust the update notice.

websupporter:
The upgrade notice was for 2.6.9. I do not think we need one for 2.7.0 since there are no breaking changes.

Issue #82, opened 2017-02-02
Broken localization in `Antispam_Bee_Columns::print_plugin_column()`


schlessera:
The method `Antispam_Bee_Columns::print_plugin_column()` contains a broken localization, where the the text to be translated is not a literal string, but a variable `$spam_reasons[ $spam_reason ]`. This will not work with `gettext`, which only does very basic string substitution.

websupporter:
Reference https://github.com/pluginkollektiv/antispam-bee/blob/master/inc/columns.class.php#L63

schlessera:
Resolved with #111

PR #111, opened 2017-02-05
Spam Reasons translateable

websupporter:
Solves #82 

The spam reasons are translateable now and I quickly changes the name "RegExp" to "Regular Expression".


antispam_bee.php
```diff
@@ -396,7 +396,7 @@ private static function _init_internal_vars()
 				'dnsbl'		=> __( 'DNSBL Spam', 'antispam-bee' ),
```
schlessera: If you are already making changes to the names, the DNSBL was an old spam database that we don't use anymore. This now uses StopForumSpam database. Maybe we should even change the key `dnsbl` and the method names to keep it consistent, not only the string.

schlessera: Hmm, but maybe in a different PR...

websupporter: Could have been seperated. I agree. When we change the key, we get a problem with old data, where the spam reason is saved as `dnsbl`, which will not be recongnized no more, when I see this correctly?

schlessera: Yes, we need a migration path for spam reasons, if they evolve like that

websupporter:
For now, I use a more generic term for DNSBL: `Public Antispam DB`, its already quite long, so I didn't wrote `database`.

websupporter:
@schlessera maybe you can have a look.

1. Why did it fail :laughing: 
2. Is this the right place for stub functions?

schlessera:
Was an unrelated error with `rake` build failing. I restarted the check to see if it happens again...

websupporter:
Maybe its because the branch does not have the yaml, since this file was introduced later?

schlessera:
Yes, that might be the cause. Try rebasing the branch.

PR #118, opened 2017-02-06
add the .travis.yml

websupporter:
Adds a basic travis.yml

Since we support PHP from 5.2x on, I've PHP support from 5.2 upwards.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,14 @@
 language: php
 
 php:
-  - '5.2'
-  - '5.3'
-  - '5.4'
+#  - '5.2'
+#  - '5.3'
+#  - '5.4'
   - '5.5'
   - '5.6'
   - '7.0'
   - '7.1'
-  - hhvm
-  - nightly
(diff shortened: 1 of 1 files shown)


websupporter:
Since our test environment needs at least 5.5, we unfortunatly can't test below this PHP version.

websupporter:
It looks like, we do not support HHVM right now. Since we never claimed, I remove this test for now. (see the result in https://travis-ci.org/pluginkollektiv/antispam-bee/jobs/198738611)

Same goes with the PHP nightly version (https://travis-ci.org/pluginkollektiv/antispam-bee/jobs/198738612), we do use functions, which will be `depreceated` in version 7.2, I think this is good to know and should be kept in the tests.

websupporter:
I also uncommented nightly for now (basically to see if we get a clean test now).

schlessera:
Merged with 5.5 for now. We should work on lowering the test requirements (non-blocking).

Issue #121, opened 2017-02-08
Allow comma separated values in blacklist/whitelist


Zodiac1978:
Using a comma separated list of ISO codes ...
![bildschirmfoto 2017-02-08 um 15 22 18](https://cloud.githubusercontent.com/assets/870475/22741207/a4d0cbe4-ee12-11e6-846e-c318e4495581.png)
has this result:
![bildschirmfoto 2017-02-08 um 15 22 32](https://cloud.githubusercontent.com/assets/870475/22741208/a4d9e814-ee12-11e6-88b4-9125e26cc980.png)
I don't think this is the expected result. I Would expect a comma separated list (like in the placeholder). Omitting the space should not have this result and we shouldn't remove the comma.

websupporter:
Needs to be fixed. Will do it this evening.

schlessera:
Resolved with #122

PR #122, opened 2017-02-08
fix regex for saving country data

websupporter:
Solves #121 

We did adjust the regex to split the country list string in _antispam_bee.php_ to `/[\s,;]+/` in #86 and changes the placeholder accordingly in #87. The idea was to be able to use comma or semicolon as well to seperate the country codes.

However, we have missed, the rexeg sanitizing the input in _inc/gui.class.php_, which still just allowed A-Z as well as space. This PR extends the allowed characters to comma and semicolon: `/[^A-Z ,;]/`

> I Would expect a comma separated list (like in the placeholder)

It is not exactly the output like the format of the placeholder, you can also use just space or semicolons. But I think its good enough, if we see the placeholder more as a guide, while we actually provide more possibilities to save the data correctly in order to reduce errors. But you could argue, we should format different allowed inputs into the same output string, like `DE UK, US; DK` as input should be saved as `DE, UK, US, DK`. For me this doesn't look necessary, but we could go this route.

### DIFF
--- a/inc/gui.class.php
+++ b/inc/gui.class.php
@@ -99,7 +99,7 @@ class Antispam_Bee_GUI extends Antispam_Bee {
 		// Blacklist clean
 		if ( !empty($options['country_black']) ) {
 			$options['country_black'] = preg_replace(
-				'/[^A-Z ]/',
+				'/[^A-Z ,;]/',
 				'',
 				strtoupper($options['country_black'])
 			);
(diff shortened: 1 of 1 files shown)


Zodiac1978:
I'm fine with any delimiter (space, comma, semicolon or multiple) as long it is working ;)

> For me this doesn't look necessary, but we could go this route.

For me it would be the best solution, if this clean up would be the case. But this is totally just me. How much work do you need to do this? How does this relate to the actual benefit for the users? If it does not relate, see first line ;)

PR #124, opened 2017-02-13
Fix typo.

tfrommen:
There was a stray/double `s`. Now it's gone. _Magic_!


websupporter:
Thanks for catching this @tfrommen

PR #134, opened 2017-02-20
remove usage of get_the_title()

### LINKED ISSUE #131: Comment empty
Related: https://wordpress.org/support/topic/comments-not-working-after-update/

Some reports say, Antispam Bee doesn't work no more, and you run into the WP message the comment field was left empty.

One part of this problem is caching. You will need to flush the cache to run 2.7.0 properly. But this didn't solve it for every reporter. We do have a problem in our secret generation as we rely on `get_the_title()` which provides an filter to manipulate the title. We need to be sure our comment-field-name-generation outputs the exact same outcome in the `init`-Filter and the `template-redirect`-action, but a filter flow like the following is not unlikely:

```
add_action( 'init', function() {
    add_filter( 'get_the_title', function( $title ) {
        return preg_replace( 'this', 'bythat', $title );
    }
}
```

To stabilize the outcome, we should switch to the posts ID in generating the secrets.

Furthermore, I want to investigate, what chars can be the real outcome of `sha1`. If it can be for example an `"` this might produce a problem as well. So, it might be a good idea to make sure to return only `a-z0-9` from the functions.

I will prepare an PR quickly.

websupporter:
Removes `get_the_title()` to generate the secrets.

Fixes #131.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -155,7 +155,7 @@ For the complete changelog, check out our [GitHub repository](https://github.com
 == Upgrade Notice ==
 
 = 2.7.1 =
-Please flush your cache, after you have updated.
+This update changes the actual markup of your comment fields. Please make sure that you flush your caches after the plugin was updated!
 
 ## Screenshots ##
 1. Antispam Bee settings
--- a/readme.txt
+++ b/readme.txt
@@ -152,5 +152,10 @@ A complete documentation is available in the [GitHub repository Wiki](https://gi
 
 For the complete changelog, check out our [GitHub repository](https://github.com/pluginkollektiv/antispam-bee).
 
+== Upgrade Notice ==
+
+= 2.7.1 =
+Please flush your cache, after you have updated.
+
 ## Screenshots ##
 1. Antispam Bee settings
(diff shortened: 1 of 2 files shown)


websupporter:
Solves #131

websupporter:
If we use this PR, we have to add the change notice to flush the cache!

Zodiac1978:
This is causing problems in the wild, so we should merge this for 2.7.1 and like I write in #139 - we should add this upgrade notice

readme.txt
```diff
@@ -152,5 +152,10 @@ A complete documentation is available in the [GitHub repository Wiki](https://gi
 
 For the complete changelog, check out our [GitHub repository](https://github.com/pluginkollektiv/antispam-bee).
 
+== Upgrade Notice ==
+
+= 2.7.1 =
+Please flush your cache, after you have updated.
```
schlessera: I find this sentence to bee a bit clumsy, mainly because of the comma.

How about:
```
This update changes the actual markup of your comment fields. Please make sure that you flush your caches after the plugin was updated!
```

websupporter: Done :)

Issue #132, opened 2017-02-19
Add more languages to Translate API Check


Zodiac1978:
This is an feature request from the support forums:
https://wordpress.org/support/topic/adding-allowed-language/

As @krafit has added to the PR, there are more language possible:
https://cloud.google.com/translate/docs/languages

Maybe we could add an filter to make it easier to change?

schlessera:
Fixed through #135

PR #135, opened 2017-02-20
add hebrew, make allowed languages filterable (solves #132)

websupporter:
This PR adds Hebrew to the languages and adds a filter, so you can easily add more languages to be selectable and solves #132

I am not quite sure, if we should go and add all possible languages (as a bad UX might outweigh the benefits) or restrict ourselves to the most spoken ones (or the ones, which get requested). Happy to add more languages.

### DIFF
--- a/inc/gui.class.php
+++ b/inc/gui.class.php
@@ -83,7 +83,9 @@ class Antispam_Bee_GUI extends Antispam_Bee {
 
 		// Translate API
 		if ( !empty($options['translate_lang']) ) {
-			if ( !preg_match('/^(de|en|fr|it|es)$/', $options['translate_lang']) ) {
+			$lang = self::get_allowed_translate_languages();
+			$lang = array_keys( $lang );
+			if ( ! in_array( $options['translate_lang'], $lang, true ) ) {
 				$options['translate_lang'] = '';
 			}
 		}
@@ -365,7 +367,9 @@ class Antispam_Bee_GUI extends Antispam_Bee {
 								<ul>
 									<li>
 										<select name="ab_translate_lang">
-											<?php foreach( array('de' => 'German', 'en' => 'English', 'fr' => 'French', 'it' => 'Italian', 'es' => 'Spanish') as $k => $v ) { ?>
+											<?php
+											$lang = self::get_allowed_translate_languages();
+											foreach( $lang as $k => $v ) { ?>
 												<option <?php selected($options['translate_lang'], $k); ?> value="<?php echo esc_attr($k) ?>"><?php esc_html_e($v, 'antispam-bee') ?></option>
 											<?php } ?>
 										</select>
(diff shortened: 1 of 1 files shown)


Zodiac1978:
Would it make sense to use the UI from the blacklist/whitelist check here again?

Without using something like select2 I think the list would be too long for all entries. Or not?

krafit:
Manually adding a ISO Code seems like a rather bad UX, dosen't it?
I do not really like the whitelist/blacklist UI.

How about our current dropdown with 5 most used languages *PLUS* (if not included in those 5) the current locale, if covered by the Google Translation API? This way we'll cover most cases (based upon the assumption you expect comments in the language your site is written in) while still maintaining a slim select-field. 
Edge cases with other languages will still be covered by the filter.

Zodiac1978:
> How about our current dropdown with 5 most used languages PLUS (if not included in those 5) the current locale, if covered by the Google Translation API?

That's a great idea. Slim select plus sane defaults. 

@websupporter Can you make this possible?

schlessera:
I had done a select box for a client where the list of elements could potentially be huge, but in 95% of the cases, only a very limited selection did make sense.

I created a select box that was split into two sections, with the most probable selections in the first section (for easy clicking, and then the rest of the long list of selections below in the second section. Of course, one could just type to get to any selection quickly.

See mockup here: http://codepen.io/schlessera/pen/VeoRmY

schlessera:
This is the component I used: http://selectize.github.io/selectize.js/

schlessera:
Ah, btw, that component of course allows for multiple selections as well.

krafit:
This _advanced_ select, @schlessera suggests looks really cool. 
But I believe we should stick with standard controls as long as possible. It's closer to the way the core handles things, and dose not require on additional (even though very little) JS.

websupporter:
Sorry for replying so late. 

Basically my approach would be: Go slim. What I mean. I am thinking about @krafit suggestion. It needs quite some work and thought to match the different locales together, which introduces for me a lot of possible bugs. For me 2.7.1 is basically to close bugs. I know, this PR is more a feature request, but basically its just a "please add a filter". If we keep it more or less on this level, for me it would be in scope for 2.7.1, otherwise we should take it into another milestone, so we do not slow down in releasing our bugfixes for 2.7.0

My prefered way would be: Hey lets add an filter in 2.7.1 and give it more thought on how we can improve the whole feature, as @schlessera suggests: We do miss multiselect. What about blogs writing in two languages, and there are other reasons to allow more than one language. 

Blacklisting or Whitelisting would also be interesting, but for me, this would need quite some UI thought.

Now, it looks a bit arbitrary to me just adding Hebrew. It was "lets go, it was requested, put it in". But maybe we should just keep it to the filter?

krafit:
jup, @websupporter is absolutely right. A enhanced select field (of any kind) should not be part of the `2.7.1` milestone. And adding Hebrew to our set of available options seems no right.

Adding a filter (and a paragraph to the wiki) will do the job just fine for now.

Zodiac1978:
Although I really like the idea of a filter we should have in mind, that people do not read everything carefully. If we introduce a filter now it should still work with the new system. If not things will break or not work anymore wit the next release, which could lead to support questions / complaints / etc.

websupporter:
The filter enables you to add or remove available languages. The out-coming array is used to build the select and to check whether the selected value is valid. I think the filter itself is quite forward compatible. If we switch to multiselect, we still need an array of available languages. If we switch to a blacklist/whitelist something, we still need the same array. So I do not see any problem here. The only thing, this filter might almost  get obsolet would be, if we add all available languages, Google has. In this case the only use cases would be to remove a language or to change the languages name. But right now, I can't imagine a scenario, where we would need to get rid of this filter.

websupporter:
The last commit removes Hebrew. If its needed it can be added via the filter:
```
add_filter( 'ab_get_allowed_translate_languages', function( $languages ) {
  $languages['iw'] => 'Hebrew';
  return $languages;
});
```

Zodiac1978:
@websupporter Shouldn't it be `he` for Hebrew?
https://en.wikipedia.org/wiki/Hebrew_language

I'm perfectly fine with merging this for now (2.7.1) and elaborate things in 2.8, maybe with the idea from @krafit

websupporter:
It's `iw` in ISO-639-1 according to 
see https://cloud.google.com/translate/docs/languages

Zodiac1978:
Strange, according to Wikipedia it is `he`:
https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes

websupporter:
```

PR #137, opened 2017-02-20
Disable autocomplete on comment input field.

### LINKED ISSUE #130: Chrome autofill seems to mistake the comment field for an address field
On some sites, Chrome Autofill mistakenly fills out the `comment` field as if it where a street address field, as reported here: https://wordpress.org/support/topic/merkwurdiges-verhalten/#post-8812548

I suggest adding hints for browser so that they can make out what the different form fields mean, and set the actual comment field to `autofill off`.

Link to spec: https://html.spec.whatwg.org/multipage/forms.html#autofill

schlessera:
Fixes #130

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1288,7 +1288,7 @@ class Antispam_Bee {
 			$init_time_field = '';
 		}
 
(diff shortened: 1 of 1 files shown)


websupporter:
Looks good to me. Thanks a lot @schlessera

Issue #142, opened 2017-03-06
autocomplete=off does not help on Chrome


websupporter:
I had the chance to test our master branch a bit and unfortunately `autocomplete=off` is simply ignored by Chrome.

> we started ignoring autocomplete=off for Chrome Autofill data

https://bugs.chromium.org/p/chromium/issues/detail?id=468153#c164

The proposed solution:

> In cases where you really want to disable autofill, our suggestion at this point is to utilize the autocomplete attribute to give valid, semantic meaning to your fields. If we encounter an autocomplete attribute that we don't recognize, we won't try and fill it.

or as MDN puts it:
> autocomplete="nope"

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

schlessera:
Ridiculous! (Google Chrome, not you, @websupporter)

websupporter:
:joy:

PR #143, opened 2017-03-06
replace autocmplete=off with autocmplete=nope (solves #142)

websupporter:
To get over with this. This solves #142. I did chose `nope` instead of `off` as suggested here https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion.


swissspidy:
FYI You can write "Fixes #xy" in the _description_ next time to automatically close the related issue.

PR #146, opened 2017-03-19
Use correct URL for "Use a public antispam database" wiki link

timse201:


swissspidy:
In the future, please create an issue beforehand. Makes tracking things easier 🙂

PR #148, opened 2017-03-22
Extended the regex to match textareas with content.

### LINKED ISSUE #147: `preg_replace_callback` does not work if textarea is prefilled.
Example:
`<textarea name="comment">Comment here...</textarea>`

We do not catch such a `textarea` to replace it with our textarea.

@Zodiac1978 reported this for this framework: https://github.com/CherryFramework/CherryFramework/blob/master/comments.php#L73

I just got the same issue for another theme. Should we handle this edge case? So far, we concluded in Slack not to support this Edge case, but maybe it is not so rare as we figured?

2ndkauboy:
Fixes #147.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1151,7 +1151,9 @@ class Antispam_Bee {
 					(?P<between3>[^>]*)
 				)
 				(?P<after>[^>]*)                                                    (?# match any additional optional attributes )
-				><\/textarea>                                                       (?# the closing of the textarea )
+				>                                                                   (?# the closing of the textarea opening tag )
+				(?s)(?P<content>.*?)                                                (?# any textarea content )
+				<\/textarea>                                                        (?# the closing textarea tag )
 			)/x',
 			array( 'Antispam_Bee', 'replace_comment_field_callback' ),
 			$data,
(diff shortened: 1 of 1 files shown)


websupporter:
Hi @2ndkauboy,
I tested the Regex with the following string https://regex101.com/r/7Ywc6M/1

Not quite sure, but, imagine you have a sidebar following with a contact form textarea. For me it looks like this would bring problems.

2ndkauboy:
I also tried that out and I know that multiple textareas are a problem. But is this really an issue? On which content is the Regex executed? Is it really the whole HTML markup? Do can this error actually occur?

websupporter:
I've played a bit, what do you think about `(?s)(?P<content>([\s\S]*?)) ` @2ndkauboy https://regex101.com/r/7Ywc6M/2

websupporter:
A I didn't see your comment. Unfortunalty yes. Right now, we are hooked into `template_redirect`. I was thinking too, if we should change it, but I do not know, what reason was behind this hook or if we can easily switch to a more appropriate filter.

2ndkauboy:
@websupporter we both found a regex, that solved the issue of capturing more content than we want to capture, when there is a second `<textarea>`. But there was a really simple solution, that is a lot better and more performant. Instead of `.*` I now just use `.*?` to make the capture "lazy" instead of "greedy". I've tested the regex with some examples and for me it looks good to be merged.

websupporter:
Thank you @2ndkauboy and sorry this PR took so long.

PR #154, opened 2017-06-04
Use franc for natural language detection

sergejmueller:
Uses https://github.com/sergejmueller/franc-aws microservice instead of Google Translate API for
language detection.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -279,7 +279,7 @@ class Antispam_Bee {
 */
 
 	/**
-	* Action during the activation of the Plugins 
+	* Action during the activation of the Plugins
 	*
 	* @since   0.1
 	* @change  2.4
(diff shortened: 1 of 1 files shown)


websupporter:
Hi @sergejmueller,
thanks a lot for the PR. This would be cool. What worries me a bit right now is our new filter to add more languages: https://github.com/pluginkollektiv/antispam-bee/blob/master/inc/gui.class.php#L559
(related #132)

With the `_map_lang_code` method it seems to me, we would close the possibility to add new languages via the mentioned filter. https://github.com/pluginkollektiv/antispam-bee/pull/154/commits/367389a401a508a6b57d2ccf2baaf0b8bdeb55f5#diff-33ae96de74050319957e1ac65615cf29R2019

Since franc is using a different standard, we would either need a complete `ISO > Franc` mapping, or a clear cut and switch to the new format (clearly communicated in the readme).

What do you think?

Thanks a lot again :heart:

PR #181, opened 2018-04-19
remove external spam DB closes #180

### LINKED ISSUE #180: Remove public antispam database due to GDPR
GDPR is right around the corner and no matter what, the use of an external public spam DB seems like an unsolvable problem to me. The public spam db check has always been optional, but is a potential GDPR risk for uniformed users.

So I suggest its complete removal from Antispam Bee.

I don't think of the public db as a real important filter method, since ASB has a more than sufficient success rate even without it.

krafit:
Remove public spam DB due to potential GDPR violation

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -379,7 +379,6 @@ class Antispam_Bee {
 				'translate_api' 	=> 0,
 				'translate_lang'	=> '',
 
-				'dnsbl_check'		=> 0,
 				'bbcode_check'		=> 1,
 
 				// Advanced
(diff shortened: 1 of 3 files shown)


antispam_bee.php
```diff
@@ -1805,50 +1789,7 @@ private static function _is_country_spam($ip)
...
-		// Empty data
-		if ( empty($result->success) ) {
-			return false;
-		}
-
-		// Return status
-		return (bool) $result->ip->appears;
-	}
-
+
```
websupporter: Nitpicking: L1792, there is a tab or something.

krafit: good catch.... but why?! 😱 🤔

websupporter:
Can you remove the check from the readme? https://github.com/pluginkollektiv/antispam-bee/blob/master/readme.txt#L26

PR #185, opened 2018-04-19
Remove Dutch translation files

### LINKED ISSUE #184: Remove Dutch translation
The Dutch translation on [translate.wordpress.org](https://translate.wordpress.org/locale/nl/default/wp-plugins/antispam-bee) is at 100%, so it's time for us to get rid of ye ol .po/mo files :tada:

See #15 for further information.

krafit:
closes #184


websupporter:
@krafit we should keep an eye on the Russian translation though. But currently it is not 100% tranlsated into Russian

https://translate.wordpress.org/locale/ru/default/wp-plugins/antispam-bee

krafit:
@websupporter yep. I checked it. All the things we have in our files are already on .org, but as long as the remaining 45% won't get translated as well, ru_RU will not get a language file form .org 😢

PR #188, opened 2018-04-20
Introduce coding standards

websupporter:
This PR introduces WordPress coding standards to Antispam Bee

* Add coding standards and configuration to CI
* Adjust code styles where necessary (_hüstl_)

We might want to discuss, what standards actually to follow and which rules to exclude (see phpcs.config.xml), but I thought, lets get this started.


inc/columns.class.php
```diff
@@ -1,120 +1,105 @@
...
+	 * Register plugin columns on comments screen.
+	 *
+	 * @since   2.6.0
+	 * @change  2.6.0
+	 *
+	 * @param   array $columns Array with existing columns.
+	 * @return  array          Array with extended columns.
+	 */
+	public static function register_plugin_columns( $columns ) {
 		return array_merge(
```
bueltge: Before return statement the WP Codex need a empty line @websupporter

inc/columns.class.php
```diff
@@ -1,120 +1,105 @@
...
+	 * Display plugin column values on comments screen
+	 *
+	 * @since   2.6.0
+	 * @change  2.6.0
+	 *
+	 * @param   string  $column      Currently selected column.
+	 * @param   integer $comment_id  Comment ID.
+	 */
+	public static function print_plugin_column( $column, $comment_id ) {
+		if ( 'antispam_bee_reason' !== $column ) {
```
bueltge: Also here a empty line please @websupporter

inc/columns.class.php
```diff
@@ -1,120 +1,105 @@
...
 			return;
 		}
 
-		// Init data
-		$spam_reason = get_comment_meta($comment_id, $column, true);
+		$spam_reason  = get_comment_meta( $comment_id, $column, true );
 		$spam_reasons = Antispam_Bee::$defaults['reasons'];
```
bueltge: With or without whitespace in the brackets Line 48 vs. line 50; I mean WP codex need the space. @websupporter

websupporter:
Hi @bueltge,
currently the code seems to apply to the coding standards as defined in the WordPress Coding standard. This PR also introduces the travis support for this.

See: https://github.com/pluginkollektiv/antispam-bee/pull/188/commits/453a17af849aaab0730e1b47e5fad0944df3901
https://travis-ci.org/pluginkollektiv/antispam-bee/jobs/377464784#L509

>Before return statement the WP Codex need a empty line @websupporter

At least for the automatic test this seems to be okay. The question would be, (1) if you want to follow another WP coding standard (not sure, maybe the Core, Extra or VIP standard do include these rules), or (2) if we want to add those rules and alter the standard slightly.

My aim in this PR would only be to have a coding standard we can test against and a code which passes the test. What I mean, if we introduce standards, which are not covered in the tests, we only provoke to break those rules.

The problem I am having with introducing new rules and changing the coding standard, we would need to explicit them. I think, we do not have a contribute.MD yet, but once we have one, currently we could simply say "follow WP standards as defined here -> w.org/something/standards". If we start our own ruleset, we would need to document them somewhere.

What do you think?

bueltge:
We don't follow another coding standard so that we reduce the complexity for newbies there maybe also help us. The default wp standard, like storm and phpcs should enough. I think we should living a simple coding standard, defined and accepted in the community, currently the wp core standard.

websupporter:
Thanks @bueltge,
did run it locally also against the behat tests and it worked fine, so I will merge it now.

inc/gui.class.php
```diff
@@ -33,7 +33,7 @@ public static function save_changes() {
 
 		$selected_languages_raw = wp_unslash( self::get_key( $_POST, 'ab_translate_lang' ) );
 		if ( ! is_array( $selected_languages_raw ) ) {
-			$selected_languages_raw = [];
+			$selected_languages_raw = array();
```
bueltge: Which php version is necessary? is the long string version of array necessary?
Currently we set it for PHP 5.2.4; however maybe we should switch. Have we more points to switch?

websupporter: The `array()` was introduced by merging the master again back into this branch. The reason was #209 but not coding standards itself. The coding standards do not restrict us to use `array()`. Although personally I am in favour of switching to a higher PHP version, we should be careful here, as we now know there are installations out there with PHP < 5.4 and we do not really know how many installations would be affected of such a step. But it is certainly worth to think about it.

bueltge: Yes, I think also we should carefully. Only if we have sufficient points to switch in a higher version we should switch. The codex should not a point. Ok, so we had to adjourn the topic of php version for the plugin.

krafit: I don't think raising the minimum required PHP version is a step we should take in just one update. I have some thoughts on this, and might open a new issue for it :)

PR #194, opened 2018-05-10
save hashed ip in meta and compare current ip with hashes

### LINKED ISSUE #177: IP address check will produce a lot of false positives
With the GDPR coming, a lot of people will delete the IP address automatically, leaving an empty string or 127.0.0.1 or something similar. If this happens before our IP comparison, it will produce a lot of false positives. If it happens after our IP comparison, the comparison is useless.

We should consider to refactor this check.

This is, what I see right now:

1. Get rid of the check completely
2. Decouple email/IP/website-check, which is currently done in one step and give the user the ability to remove only the IP check

We should do this quite soon.

websupporter:
closes #177 

with gdpr coming up, the comment_author_ip column won't be reliable anymore as more and more people will rightfully start to anonymize this column. even core might follow up on this.

on the other hand the ip address is a very valuable source to determine spam and has been proven effective in preventing spam when stored in the local database.

this approach tries to bring privacy and spam protection together. it stores a salted/hashed version of the ip address of spam comments. if a comment is incoming, we check against this hashed ip.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1687,18 +1687,25 @@ class Antispam_Bee {
 		// Global
 		global $wpdb;
 
-		$hashed_ip = self::hash_ip( wp_unslash($ip) );
-
-		$args = array(
-			'meta_key' => 'antispam_bee_iphash',
-			'meta_value' => $hashed_ip,
-			'status' => 'spam',
-		);
-		$query = new WP_Comment_Query($args);
-
-		if ( ! empty( $query->get_comments() ) ) {
-			return true;
+		$sql = '
+			select 
+				meta_value as ip
+			from
+			 	' . $wpdb->commentmeta . ' as meta,
+			 	' . $wpdb->comments . ' as comments
+			where
+			    comments.comment_ID = meta.comment_id
+			    AND meta.meta_key = "antispam_bee_iphash"
+			    AND comments.comment_approved="spam"';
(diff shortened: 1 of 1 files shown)


antispam_bee.php
```diff
@@ -2299,6 +2333,21 @@ public static function add_spam_reason_to_comment( $comment_id )
...
+		add_comment_meta(
+			$comment_id,
+			'antispam_bee_iphash',
+			$hashed_ip
+		);
+	}
+
+	public static function hash_ip($ip)
+	{
+		return substr( sha1( md5( self::$_salt . $ip ) ), 0, 10 );
```
stklcode: Hope it's clear that this is no real anonymization, just some added obscurity. Protection against a third party it is, but not for an admin or other plugins knowing the salt.

At least for IPv4 there is a high chance for no collision within the first 10 characters (depending on the salt) and the computation is lightweight, thus the real IP can be "calculated" within seconds on a modern consumer-grade computer.

websupporter: Thanks for your feedback @stklcode, you are right. I was thinking about changing this with  `wp_hash_password()` and using `wp_check_password()` somehow. But the current implementation shouldn't go into master.

websupporter: My current approach would be

krafit:
Thank you @websupporter ❤️

Issue #208, opened 2018-05-23
Version 2.8.0 not compatible with PHP 5.3.x


benderl:
My webhoster still uses PHP in Version 5.3.x and I am not able to upgrade. After updating antispam-bee I got an error 500 when going to the antispam-bee settings. Please replace the brackets '[]' in lines 40 and 42 of gui.class.php with 'array()' and your plugin will work even with PHP 5.3.x.

https://github.com/pluginkollektiv/antispam-bee/blob/045b9f2dc3d896d87eb6369b61a8fd88e0f5eccd/inc/gui.class.php#L42

benderl:
Please have a look at #209.

websupporter:
Hi @benderl,
thanks a lot for your pull request. not used any more to `array()`, I am very sorry. This is now merged in master and will be in the upcoming version `2.8.1` which will be released soon.

Sorry again for the trouble.

PR #211, opened 2018-05-26
Introducing Behat tests

### LINKED ISSUE #210: Introducing behat tests
To see, if Antispam Bee continues to do the right thing and not just the thing right, we should introduce some behat tests to stabilize our development process

websupporter:
closes #210 

I am very happy to introduce some behat tests to Antispam Bee :tada: 

*What is covered?*
These tests cover almost all comment filters and already a bit the trackback section, which would be my next focus. They also cover a bit the admin area, but here I want to add some more tests as well.

*What gets tested?*
The tests run on WP 4.5 -> WP nightly, so a sudden surprise like we had in ASB v2.8 were we missed, we would need to raise the min version should be much less likely :tada:

*Whats the current downside*
We use testing tools ([WordHat](https://github.com/paulgibbs/behat-wordpress-extension))which require PHP 7.0 at least. Before the introduction of the tests, we were able to run our unit tests also on PHP 5.5 and PHP 5.6, this is currently not possible.

But all in all I am super excited and happy :smile:


schlessera:
@websupporter I unfortunately have no time right now to go through this in detail, due to the upcoming WCEU. I'll have more time once that is over.

As a quick feedback, in case you want to move forward with this in the mean time:

- I like having integration tests, and you did a great job incorporating them.
- I'm a bit unhappy about the high version requirement, which causes tests to only run for a minority of our "customer base" I think.
- Given the high version requirement for the integration test, we should at the very least run the PHP linter at lower versions, to detect when we use PHP language constructs at lower versions that are not yet available.

From my part, +1 on merging this to not letting it go stale, and iterating further in additional PRs.

websupporter:
We run our unit tests now from php 5.4 upwards. Basically, I added a new composer.json file, (`composer-php-5.4.json`), which is set as composer.json for PHP tests running below php 7.0. For unit tests it seems 5.4 is the lowest we can go, as our testing framework does not allow for a lower php version.

Basically, this PR does not only include behat tests but also enables us to run our unit tests now on PHP 5.4 (before the lowest was PHP 5.5)

I tried to push the code sniffer even deeper, but there it seems to stop to take our `phpcs.config.xml` into consideration while running the coding standard tests.

Basically its a pitty to not be able to cover php 5.3, considering #208. I think it would be worth to have again a look into this to be able to test below < 5.4, but for now this extends the scope of this PR.

I think for now, we are good to go, so iI am going to merge. thanks @schlessera for having an additional look into the PR.

PR #223, opened 2018-05-31
introduce blog name and post title check

### LINKED ISSUE #219: new filter for trackbacks: blog_name == title of blog post
In #213 we see a nice pattern, we could introduce as a new filter for trackbacks:

If a trackback from another blog is sent, it will be rarely the case that the name of the blog sending and the title of the blog post from which the trackback originates would be the same.

But in the pattern we see, both are equal and we could detect this.

websupporter:
closes #219 

if you send a trackback, you send the title of the blog as well as the title of the post, from which the trackback originates. We have seen, there are spam comments, where those both values are equal. This filter would qualify those trackbacks as spam.

Currently, I haven't added any settings. This filter would run for every trackback. I hesitate to add an option for this filter as

1.) We have already too many options and the users do not quite understand what they are for.
2.) I would not know, where to place this option in our current UI

Feedback welcome :smiley:


krafit:
+1 for not adding a setting for this

websupporter:
Thanks for your feedback. Any ideas about name, description, place in the UI @krafit

krafit:
oh, I would add it to the UI at all

PR #225, opened 2018-05-31
Switch language detection to franc

websupporter:
This builds upon #154 and closes #154 

I have updated the code and mapped all languages I was able to find in franc (list is here: https://github.com/wooorm/franc/tree/master/packages/franc-min) to ISO 639-1, the standard we use.

I would suggest to keep our standard to ISO 639-1, so we do not break the currently saved states and are able to switch back to a service which uses ISO 639-1. Not all languages seem to have a code in this standard though.

We introduced a filter in #135 to filter the languages. Specifically we introduced it, so a user could add Hebrew. Now: Hebrew seems not to be supported by franc. If we switch to franc, which would save quite a lot of money, we should try to reach out to this user, to find a solution.

My suggestion is the newly introduced filter `antispam_bee_detected_lang`. This would enable to add the Google API back again (example would follow)


krafit:
@websupporter actually `franc` supports Hebrew as well. There are three packages - `franc-min`, `franc` and `franc-all`. They support 82, 188, or 402 languages and Hebrew is included in second and third package.

krafit:
yep, I already looked into that, and I'd like to give it a try - running the service on our own.

antispam_bee.php
```diff
@@ -1735,53 +1735,142 @@ private static function _is_lang_spam( $comment_content ) {
...
+			'bul' => 'bg',
+			'swe' => 'sv',
+			'lin' => 'ln',
+			'som' => 'so',
+			'hms' => 'hms', // Seems not to have a ISO 639-1 code.
+			'hnj' => 'hmn', // Seems not to have a ISO 639-1 code. Google seems to use hmn.
+			'ilo' => 'ilo', // Seems not to have a ISO 639-1 code.
+			'kaz' => 'kk',
+		);
+		if ( array_key_exists( $franc_code, $codes ) ) {
```
bueltge: Missing empty line before `if`.

antispam_bee.php
```diff
@@ -1735,53 +1735,142 @@ private static function _is_lang_spam( $comment_content ) {
...
+			return $codes[ $franc_code ];
+		}
+		return $franc_code;
```
bueltge: Missing empty line before `return`.

websupporter:
The latest list basically uses the codes, I found [here](https://iso639-3.sil.org/code_tables/639/data/all?title=&field_iso639_cd_st_mmbrshp_639_1_tid=255291&name_3=&field_iso639_element_scope_tid=All&field_iso639_language_type_tid=All&items_per_page=500). 

I just quickly scraped it with the followingg code:
```
let tablesNodes = document.querySelectorAll('.views-table');
var tables = Array.prototype.slice.call(tablesNodes);
var list = [];
for( var i = 0; i < tables.length; i++ ) {
	list = Array.prototype.slice.call(tables[i].querySelector('tbody').querySelectorAll('tr')).concat(list);
};
var iso6393 = [];
for( var i = 0; i < list.length; i++ ) {
	iso6393.push(list[i].querySelector('.views-field-field-iso639-639-3-code').innerText);
}
var iso6391 = [];
for( var i = 0; i < list.length; i++ ) {
	iso6391.push(list[i].querySelector('.views-field-field-iso639-639-1-code').innerText);
}

var codes = [];

for ( var i = 0; i < iso6391.length; i++ ) {
	if( iso6391[i].length === 0 ) {
		continue;
	}
	codes.push([iso6393[i],iso6391[i]]);
}

for ( var i = 0; i < codes.length; i++ ) {
	
	var current = codes[i];
	if( !current[0].length || !current[1].length ) {
		continue;
	}

	console.log( "'" + current[0] + "' => '" + current[1] + "'," );
}
```

This way, we get >180 codes assigned. I think the most important are present (especially the ones, we use in ASB core). Also I took care of the empty lines.

websupporter:
We have to update the wiki:

```

bueltge:
@stkjj Schaue mal in die Anleitung: https://help.github.com/articles/adding-and-editing-wiki-pages-locally/
Das Repo clonen: https://github.com/pluginkollektiv/antispam-bee.wiki.git - da kannst du lokal bauen und pushen.

stkjj:
den Teil hab ich (glaube ich) schon: 

bildschirmfoto 2018-07-24 um 15 46 09

Aber das jetzt zu commiten gelingt mir nicht: 403 dürfte heissen, das ich da keine Rechte für habe?!

bueltge:
Stimmt, du bist aktuell kein Mitglied des Pluginkollektiv, damit keine Rechte. Du kannst nur Forken, Pull Request stellen. Ich will nicht einfach Leute ergänzen, kenne keine Details. Aber im Slack des Pluginkollektiv bekommst du bestimmt eine Antwort oder ein OK.

websupporter:
Thanks for everyone helping in this PR especially @sergejmueller, @stkjj, @krafit, @bueltge and @2ndkauboy . I am going to merge now :tada:

PR #233, opened 2018-06-24
delete meta data on uninstall (closes #232)

### LINKED ISSUE #232: Delete IP hashes when plugin gets deactivated
Its good practice to clean up, when you leave a guest room. Currently, we do not clear the newly introduced hashed IP address when we leave the building. I suggest, we should do so with the next minor version.

:wastebasket:

websupporter:
This PR closes #232,

If you uninstall, this PR will delete the meta data spam reason and ip hash by default. It provides also an UI and option, so you can prevent this behaviour.


websupporter:
typo in checkbox
> Delete Antispamm Bee data when uninstalling

websupporter:
typo fixed, changes added to changelog + readme. phpcs related commit from #234 cherry-picked.

PR #236, opened 2018-07-04
update raphael.js to 2.2.7

### LINKED ISSUE #231: Raphaël does not work when CSP disallows 'unsafe-eval'
The version of Raphaël used in this plugin does not work on pages with a Content Security Policy disallowing `unsafe-eval` for [script-src](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src). This was fixed in Raphaël version [2.1.2](https://github.com/DmitryBaranovskiy/raphael/blob/master/history.md#212--2015-08-26).

(see https://github.com/pluginkollektiv/statify/issues/52)

websupporter:
closes #231 

thanks @abrain :heart:


websupporter:
If you look into the failure. Its the same as we experienced in #234. The latest update of code sniffer finds some errors, which haven't been detected before. We take care of it already in 
63fc250, so this error shouldn't be a blocker for this PR as its unrelated. No other errors show up during the automated tests.

PR #237, opened 2018-07-20
make links translateable

### LINKED ISSUE #199: Make links translatable
If we make these two links translatable we could switch theme to the localized versions:

https://github.com/pluginkollektiv/antispam-bee/blob/issue-191/inc/gui.class.php#L508

and 

https://github.com/pluginkollektiv/antispam-bee/blob/issue-191/inc/gui.class.php#L511

websupporter:
closes #199


websupporter:
Thanks @chesio,
much appreciated. 2bb51a6 should take care of that issue.

PR #239, opened 2018-07-20
avoid `parse_url()`

### LINKED ISSUE #77: Avoid `parse_url`

websupporter:
closes #77
replaces #85


websupporter:
`wp_parse_url()` fails bc 2. parameter was introduced in 4.7

PR #244, opened 2018-09-20
prefix letter for secret id/name (closes #242)

### LINKED ISSUE #242: Fix compatibility issues with secret name in comment form
Antispam Bee currently adjusts the comment form to use different field names, so that those cannot as easily be figured out by spam bots. For that, it replaces the comment textarea ID and name with a generated secret.

This usually works fine, however it can cause compatibility issues with other plugins if those perform additional logic on the comment form: When the secret starts with a number and then the value of the ID or name attributes are used as properties in JavaScript, this is an issue because properties should not start with a number. A concrete example is the AMP plugin for which comment forms break because of that - every comment form where the secret starts with a number will throw a JavaScript error because of the `amp-bind` integration with the comment form it needs to handle.

I used a quick fix for myself by using the `ab_get_secret_name_for_post` filter (I was amazed there is a filter for something as granular as this! 💯), but it would help if the plugin would ensure on its own that the secrets always start with a letter.

backermann1978:
Adds a letter prefix for random comment id/name via a new static property.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -65,6 +65,12 @@ class Antispam_Bee {
 	 */
 	private static $_salt;
 
+	/**
+	 * Prefix letter for secret id/name.
+	 *
+	 * @var string
+	 */
+	private static $_secret_prefix;
 
 	/**
 	 * The spam reason.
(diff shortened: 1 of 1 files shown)


antispam_bee.php
```diff
@@ -374,6 +380,8 @@ private static function _init_internal_vars() {
 		$salt        = defined( 'NONCE_SALT' ) ? NONCE_SALT : ABSPATH;
 		self::$_salt = substr( sha1( $salt ), 0, 10 );
 
+		self::$_secret_prefix = chr(rand(97,122));
```
websupporter: I think, this would mean, with every load the prefix could be different. So it could be `a` when the user submits the form and `b` after he submitted and ASB tries to determine the fields name

websupporter: strangley, it doesnt fail :joy: :thinking:

PR #249, opened 2018-10-19
Switch dashboard tooltip lines and use long date format

### LINKED ISSUE #248: Inconsistent dashboard tooltips
I currently use the Antispam Bee dashboard beside the Statify dashboard and could recognize an inconsistency in the dashboard tooltips of them.

The current tooltip of Antispam Bee:

bildschirmfoto 2018-10-18 um 12 14 28

The current tooltip of Statify:

bildschirmfoto 2018-10-18 um 12 14 34

As you can see, Statify uses the long date version and the date before the actual data. I would recommend to use the same format for Antispam Bee.

So, instead of:
```
6× Spam
18.10.2018
```
I would suggest the following:
```
18. Oktober 2018
6 × Spam
```

Additionally to the row swap I would add an additional space between the number and the `×` sign as this would be typographically correct.

KittMedia:
Will fix #248


websupporter:
Thank you @KittMedia . closes #248

PR #252, opened 2018-10-21
Set tested up to version to 5.0

florianbrinkmann:
Tested it with Gutenberg and everything seems to work (tested saving of options and submitting a comment loggged in and logged out).


websupporter:
The automatic tests also do not indicate any problems.

websupporter:
Thanks for checking @florianbrinkmann @krafit :heart:

Issue #276, opened 2019-02-22
Tests fail because PHP nightly is 8 now


websupporter:
![grafik](https://user-images.githubusercontent.com/6458412/53248187-cd253f80-36bd-11e9-8e7b-cf196e9a28bd.png)

Our dependencies do not support PHP8 yet.

websupporter:
closed via #279

PR #277, opened 2019-03-19
Update links from Google+ and old website to GitHub wiki

2ndkauboy:
I've created a `news` branch in the wiki with the archived Google+ pages. When the wiki branch is merged, this PR would link to the new news pages. I also updated some links that pointed to the old website to the matching wiki pages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -153,14 +153,14 @@
 * **English**
    * Removal of functions *Block comments from specific countries* and *Allow comments only in certain language* for financial reasons
 * **Deutsch**
-   * Entfernung der Funktionen *Kommentare nur in einer Sprache zulassen* und *Bestimmte Länder blockieren bzw. erlauben* aus finanziellen Gründen - [Hintergrund-Informationen](https://plus.google.com/u/0/+SergejMüller/posts/ZyquhoYjUyF)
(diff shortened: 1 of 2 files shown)


websupporter:
Looks good to me, the failing test is because of #276. I didn't really follow up the discussions moving die documents from G+. I guess we merge, once the moval has taken place?

Zodiac1978:
@2ndkauboy All those Github links do not work at the moment. Can you please create a PR for it or point me to the right place to do this on my own?

2ndkauboy:
The wiki has it's own Git repository and in it there is a `news` branch we would have to merge in order to get those links working.

On the hackathon someone was mentioning, that we might want to get rid of the wiki alltogether and only use the website. In that case, this PR would not make sense anymore.

Should I still merge the `news` branch as a temporary solution?

Zodiac1978:
> On the hackathon someone was mentioning, that we might want to get rid of the wiki alltogether and only use the website. In that case, this PR would not make sense anymore.

@krafit Can you she dsome light on this question?

krafit:
I incorporated @2ndkauboy Wonderfull work of saving the old Google+ posts into Antispam bees new website - https://antispambee.pluginkollektiv.org/news.
All according links were update in #327.

PR #282, opened 2019-06-03
Autoptimize compatibility

Zodiac1978:
Add data-noptimize attribute to prevent optimization from Autoptimize.


websupporter:
thanks a billion @Zodiac1978,
looks good to me.

PR #284, opened 2019-06-20
Do not store ip hash

### LINKED ISSUE #275: The IP Hash is not used but still saved
https://github.com/pluginkollektiv/antispam-bee/blob/518a2c1e4df2f597d309d83ee3dd89f58880c7be/antispam_bee.php#L268-L276

:man_facepalming:

websupporter:
closes #275

The `save_ip_hash()` method is removed and the functionality dehooked.
This PR raises the DB version of ASB, so we delete the data, we do not need anymore.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -266,17 +266,6 @@ class Antispam_Bee {
 				)
 			);
 
-			// Save IP hash, if comment is spam.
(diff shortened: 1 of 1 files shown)


pfefferle:
+1

PR #291, opened 2019-10-29
Change placeholder from integer (%d) to string (%s) - fixes #290

### LINKED ISSUE #290: Wrong spam statistics in dashboard
I stumbled upon an interesting bug. 

I have 48719 spam comments blocked so far. This is what is in my database.

But the dashboard counter says I have 48 blocked. This *little* difference made me curious. What is happening here?

But everything is working fine. If I output the counter it says correctly "48.719". Wait!? Why is there a period?

Well because of this line of formatting:
https://github.com/pluginkollektiv/antispam-bee/blob/e91ea009268516908d6d2242cb31be98aee1aa27/antispam_bee.php#L2608

So what happens in the dashboard widget? There is this placeholder `%d`:
https://github.com/pluginkollektiv/antispam-bee/blob/e91ea009268516908d6d2242cb31be98aee1aa27/antispam_bee.php#L699

And this placeholder %d means this value is an integer. The integer of "48.719" is "48".

Problem solved! What's the solution? We just need to change this to `%s` for string (there is also `%f` for float, but this is adding zeros at the end and is semantically wrong here).

PR is coming soon.

Zodiac1978:


websupporter:
:heart: Thank you @Zodiac1978

PR #294, opened 2019-11-08
Change autocomplete=nope to autocomplete=new-password (fixes #280)

### LINKED ISSUE #280: autocomplete="nope" breaks html validation - may not work anymore anyway?
Problem: The autocomplete="nope" workaround, for the issue that Chrome ignores autocomplete="off", has the unfortunate side effect, that it throws an error in w3c.org's html validation service.

While you and I know that it is probably safe to ignore that error message, teaching co-workers to "just ignore errors" is a road its maybe not wise to go down :)

Furthermore, there are rumours that "... it seems the latest version of Chrome is ignoring autocomplete attribute regardless of what your using nope or `off."
https://github.com/algolia/autocomplete.js/issues/247#issuecomment-444133128

And oh boy there are many creative (and ugly) workarounds involving jquery and whatnot :o

The most sensible solution I saw, was creating a unique random name for the field, and it seems that Antispam Bee already does that?

Another suggested setting the value attribute to a blank space, as chrome shouldn't fill out fields that already contains a value. Arguments that, that isn't a very pretty approach,forcing the user to remove a space before entering content, may not apply to a hidden honeypot field filled out by bots anyway?

If autocomplete="nope" is essential for Antispam Bee to work, last resort could be a note/warning in the backend, that this particular html validation error can and should be ignored as it's a required part of the antispam-mechanism by design.

On my personal blog, I changed the autocomplete="nope" to "off" for a succesful validation. Commenting still works. I'll monitor it over the next weeks to see if the change affects my spam amount.

Kind regards
Bjarne

Zodiac1978:
Using "new-password" validates as correct HTML5 and is preventing the autocomplete in our tests.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1195,7 +1195,7 @@ class Antispam_Bee {
 			$init_time_field = '';
 		}
 
-		$output = '<textarea autocomplete="nope" ' . $matches['before1'] . $matches['before2'] . $matches['before3'];
+		$output = '<textarea autocomplete="new-password" ' . $matches['before1'] . $matches['before2'] . $matches['before3'];
 
 		$id_script = '';
 		if ( ! empty( $matches['id1'] ) || ! empty( $matches['id2'] ) ) {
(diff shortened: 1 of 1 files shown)


websupporter:
looks good. thank you.

PR #296, opened 2019-11-08
Changing the order of IP checks (fixes #286)

### LINKED ISSUE #286: get_client_ip may fail if HTTP_FORWARDED doesn’t contain an IP address

```
Array
(
    [REMOTE_ADDR] => 127.0.0.1
)
```

Zodiac1978:
Since the HTTP_FORWARDED doesn’t necessarily contains an IP address, the function get_client_ip may fail if there is just a HTTP_FORWARDED and the REMOTE_ADDR with the following values: [REMOTE_ADDR] => 127.0.0.1 and [HTTP_FORWARDED] => proto=https. Therefore REMOTE_ADDR is moved above the HTTP_FORWARDED checks.


websupporter:
closing after quick internal discussion in favor of #297.
thank you.

PR #297, opened 2019-11-08
Improved retrieving Client-IP - Fixes #286

### LINKED ISSUE #286: get_client_ip may fail if HTTP_FORWARDED doesn’t contain an IP address

PhilWSRN:
Remote-IP is no longer used as IP when there's a valid IP from Server-HTTP-Resources.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -2361,16 +2361,7 @@ class Antispam_Bee {
 			$ip = wp_unslash( $_SERVER['HTTP_FORWARDED_FOR'] );
 		} elseif ( isset( $_SERVER['HTTP_FORWARDED'] ) ) {
 			$ip = wp_unslash( $_SERVER['HTTP_FORWARDED'] );
-		} 
-		/*
-		elseif ( isset( $_SERVER['REMOTE_ADDR'] ) ) {
-			$ip = wp_unslash( $_SERVER['REMOTE_ADDR'] );
-		} else {
-			return '';
 		}
-		*/
-
-		// phpcs:enable WordPress.VIP.ValidatedSanitizedInput.InputNotSanitized
 
 		preg_match_all ('/(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})/g', $ip, $matches );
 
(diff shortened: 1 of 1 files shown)


antispam_bee.php
```diff
@@ -2359,12 +2361,17 @@ public static function get_client_ip() {
 			$ip = wp_unslash( $_SERVER['HTTP_FORWARDED_FOR'] );
 		} elseif ( isset( $_SERVER['HTTP_FORWARDED'] ) ) {
 			$ip = wp_unslash( $_SERVER['HTTP_FORWARDED'] );
-		} elseif ( isset( $_SERVER['REMOTE_ADDR'] ) ) {
+		}
+
+		preg_match_all ('/(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})/g', $ip, $matches );
```
websupporter: does not match ipv6,
`Unknown modifier 'g'`

pfefferle: what about `filter_var($ip, FILTER_VALIDATE_IP)`?

websupporter:
Hi @PhilWSRN,
thanks for your PR, will you continue to work on it?

antispam_bee.php
stklcode: Also does not match short notations in v4, but it matches invalid records like ### _345.678.987.654_.

Either split, trim and filter separately, e.g. using PHP's built-in filter or use a better RegEx.

Also care about delimiters when using global mode patterns.  _:ffff:192.0.2.1_ is a valid address (IPv6 mapped IPv4) and must not be split in two.

PhilWSRN:
Hey Ladies & Gents,

i've been very busy since a time and will be a bit more, i'm sorry that i didn't take the time to comment on this one earlier! :/
Will contribute and continue work (with better output :P) asap.
Thanks for assigning this one to yourself @websupporter , you're in my good books!

websupporter:
closing this one in favor of a new PR as I can't directly push into Phils repo.

PR #299, opened 2019-11-08
Add a new pattern

Zodiac1978:
Add luxurybrandsale


websupporter:
Added a test. Good to go.

PR #305, opened 2019-11-11
Fix the orderby for comment spam reasons

### LINKED ISSUE #303: Ordering the comments by spam reason does not work

2ndkauboy:
Closes #303.

### DIFF
--- a/inc/columns.class.php
+++ b/inc/columns.class.php
@@ -75,17 +75,19 @@ final class Antispam_Bee_Columns {
 	 * @since   2.6.3
 	 * @change  2.6.3
 	 *
-	 * @param   \WP_Query $query  Current WordPress query.
+	 * @param   \WP_Comment_Query $query  Current WordPress query.
 	 */
 	public static function set_orderby_query( $query ) {
-		$orderby = $query->get( 'orderby' );
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended -- Input is used safely.
+		$orderby = isset( $_GET['orderby'] ) ? sanitize_text_field( wp_unslash( $_GET['orderby'] ) ) : '';
 
 		if ( empty( $orderby ) || 'antispam_bee_reason' !== $orderby ) {
 			return;
 		}
 
-		$query->set( 'meta_key', 'antispam_bee_reason' );
-		$query->set( 'orderby', 'meta_value' );
+		// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_key -- Necessary Meta_Query
+		$query->query_vars['meta_key'] = 'antispam_bee_reason';
+		$query->query_vars['orderby']  = 'meta_value';
 	}
 
 	/**
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -249,7 +249,7 @@ class Antispam_Bee {
 						)
 					);
 					add_action(
-						'pre_get_posts',
+						'pre_get_comments',
 						array(
 							'Antispam_Bee_Columns',
 							'set_orderby_query',


inc/columns.class.php
```diff
@@ -75,17 +75,17 @@ public static function register_sortable_columns( $columns ) {
 	 * @since   2.6.3
 	 * @change  2.6.3
 	 *
-	 * @param   \WP_Query $query  Current WordPress query.
+	 * @param   \WP_Comment_Query $query  Current WordPress query.
 	 */
 	public static function set_orderby_query( $query ) {
-		$orderby = $query->get( 'orderby' );
+		$spam_reason = isset( $_GET['orderby'] ) ? wp_unslash( sanitize_text_field( $_GET['orderby'] ) ) : '';
```
derpixler: Prüfung auf empty ist sicherer + evl. kürzen & Redundanz vermeiden.

```php
$orderByQueryKey = 'orderby';
$spam_reason = wp_unslash( sanitize_text_field( (!empty($_GET[$orderByQueryKey]) ? $_GET[$orderByQueryKey] : '') ) );
```
More readable 
```php
$orderByQueryKey = 'orderby';
$orderByQueryVal = !empty($_GET[$orderByQueryKey]) ? $_GET[$orderByQueryKey] : '';
$spam_reason = wp_unslash( sanitize_text_field( $orderByQueryVal ) );
```

and just a bit fancy
```php
$orderByQuery = [
	'key' => 'orderby',
	'val' => !empty($_GET[$orderByQueryKey]) ? $_GET[$orderByQueryKey] : ''
];
$spam_reason = wp_unslash( sanitize_text_field( $orderByQuery['val'] ) );
```

Zodiac1978: Why is `empty` safer than `isset`? Can you elaborate on that?

Zodiac1978: Found a useful explanation here: https://www.virendrachandak.com/techtalk/php-isset-vs-empty-vs-is_null/

derpixler: in this case `$_GET['orderby']` can be set but empty and !isset is true.

websupporter: The benefit of `!empty()` instead of `isset()` would be, you would not execute `wp_unslash()` and `sanitize_text_field()`. Other than that, I see no difference in this case.

websupporter:
Quick codesniffer:
```
$ ./vendor/bin/phpcs ./inc ./antispam_bee.php --standard=./phpcs.config.xml

FILE: /var/www/html/antispambee/wp-content/plugins/antispam-bee/inc/columns.class.php
--------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AND 4 WARNINGS AFFECTING 3 LINES
--------------------------------------------------------------------------------------------------------------
 81 | WARNING | [ ] Processing form data without nonce verification.
 81 | WARNING | [ ] Processing form data without nonce verification.
 81 | ERROR   | [ ] Missing wp_unslash() before sanitization.
 87 | WARNING | [ ] Detected usage of meta_key, possible slow query.
 88 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 2 spaces but found 1 space
--------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------------------------------------

Time: 4.86 secs; Memory: 34MB
```

Will have a deeper look this weekend.

inc/columns.class.php
2ndkauboy: I also don't see how `empty` would be more secure. I usually always use `isset` as recommended in the WPCS wiki: https://github.com/WordPress/WordPress-Coding-Standards/wiki/Fixing-errors-for-input-data#sanitization

But I just realize that I have the two functions in the wrong order (which would still work) :/

derpixler: In this case it will work.
_sanitize_text_fields will handle that. 
```php
function _sanitize_text_fields( $str, $keep_newlines = false ) {
	if ( is_object( $str ) || is_array( $str ) ) {
		return '';
	}
``` 

But if you have a case or stupid code how not prevent the error if ` $_POST['foo']` empty or null this will trigger a `undefined index` notice.

derpixler: The point of `empty()`It doesn't throw a warning if a variable does not exist or is null.

2ndkauboy: The Superglobal `$_GET` will never be null. Therefore `isset` will also not throw an error, if `$_GET['orderby']` is not set.

2ndkauboy:
@websupporter I've whitelisted two PHPCS issues and fixed the other ones.

websupporter:
Thanks @2ndkauboy,
one last question: Do we regard it as a problem that this will not necessarily mean a correct alphabetical order? What I mean is:
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L418-L429

You could expect the **C**omment Language do show up after the **H**oneypot, but actually you sort **c**ss (= honeypot) and **l**ang (=comment language)

2ndkauboy:
Unfortunately we cannot order by the translated value. But at least now we have "an order" ;)

PR #306, opened 2019-11-11
Allow filtering the comments by the spam reason

### LINKED ISSUE #302: Allow filtering the spam comments by the spam reason
Some filters may produce a high number of false positives. In order to approve them, it would help to be able to filter the spam comments list by the spam reason.

2ndkauboy:
Closes #302.


inc/columns.class.php
```diff
@@ -88,6 +88,45 @@ public static function set_orderby_query( $query ) {
...
+	 * Filter comments by the spam reason
+	 *
+	 * @global \wpdb $wpdb
+	 */
+	public static function filter_columns() {
+		global $wpdb;
+		?>
+		<label class="screen-reader-text" for="filter-by-comment-spam-reason"><?php _e( 'Filter by spam reason' ); ?></label>
+		<select id="filter-by-comment-spam-reason" name="comment_spam_reason">
+			<option value=""><?php _e( 'All spam reasons' ); ?></option>
```
Zodiac1978: Looks like the text domain is missing here

inc/columns.class.php
```diff
@@ -88,6 +88,45 @@ public static function set_orderby_query( $query ) {
...
 
+	/**
+	 * Filter comments by the spam reason
+	 *
+	 * @global \wpdb $wpdb
+	 */
+	public static function filter_columns() {
+		global $wpdb;
+		?>
+		<label class="screen-reader-text" for="filter-by-comment-spam-reason"><?php _e( 'Filter by spam reason' ); ?></label>
```
Zodiac1978: Looks like the text domain is missing here

antispam_bee.php
```diff
@@ -255,6 +255,20 @@ public static function init() {
 							'set_orderby_query',
 						)
 					);
+					add_action(
```
derpixler: Was meint Ihr, dieser gesamte Block lässt sich doch vereinfachen?
https://github.com/pluginkollektiv/antispam-bee/blob/302-filter-by-spam-reason/antispam_bee.php#L220-L271

*Beispiel:*
```
$edit_comments_hooks = [
	'filter' => [
		[
			'manage_edit-comments_columns' => 'register_plugin_columns',
			'manage_comments_custom_column' => 'print_plugin_column',
			'admin_print_styles-edit-comments.php' => 'print_column_styles',
			'manage_edit-comments_sortable_columns' => 'register_sortable_columns',
		]
	],
	'action' => [
			'pre_get_posts' => 'set_orderby_query',
			'restrict_manage_comments' => 'filter_columns',
			'pre_get_comments' => 'filter_by_spam_reason',
	]
];

foreach($edit_comments_hooks as $type => $edit_comments_hook ){
	foreach($edit_comments_hook as $hook => $val ){
		$args = ['Antispam_Bee_Columns', $val];
	
		if($type === 'filter'){
			add_filter($hook, $args);
		}else if($type === 'filter'){
			add_action($hook, $args);
		}
	}
};
```

antispam_bee.php
```diff
@@ -255,6 +255,20 @@ public static function init() {
...
+						'restrict_manage_comments',
+						array(
+							'Antispam_Bee_Columns',
+							'filter_columns',
```
derpixler: Warum haben die Filter & Action keine Prio?

websupporter: If you don't set a priority, it defaults to `10`, so I do not see a problem there.

derpixler: its a bit more consistent but i'm fine.

websupporter:
```

**Comment by websupporter** [comment]
```
$ ./vendor/bin/phpcs ./inc ./antispam_bee.php --standard=./phpcs.config.xml
 
FILE: /var/www/html/antispambee/wp-content/plugins/antispam-bee/inc/columns.class.php
--------------------------------------------------------------------------------------
FOUND 6 ERRORS AND 9 WARNINGS AFFECTING 10 LINES
--------------------------------------------------------------------------------------
  99 | ERROR   | [ ] All output should be run through an escaping function (like
     |         |     esc_html_e() or esc_attr_e()), found '_e'.
 101 | ERROR   | [ ] All output should be run through an escaping function (like
     |         |     esc_html_e() or esc_attr_e()), found '_e'.
 103 | WARNING | [ ] Processing form data without nonce verification.
 103 | WARNING | [ ] Processing form data without nonce verification.
 103 | ERROR   | [ ] Missing wp_unslash() before sanitization.
 104 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 5
     |         |     spaces but found 1 space
 108 | ERROR   | [ ] All output should be run through an escaping function (see the
     |         |     Security sections in the WordPress Developer Handbooks), found
     |         |     '">$label</option>\n"'.
 121 | WARNING | [ ] Processing form data without nonce verification.
 121 | WARNING | [ ] Processing form data without nonce verification.
 121 | ERROR   | [ ] Missing wp_unslash() before sanitization.
 122 | WARNING | [ ] Not using strict comparison for in_array; supply true for third
     |         |     argument.
 126 | WARNING | [ ] Detected usage of meta_key, possible slow query.
 126 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 3
     |         |     spaces but found 1 space
 127 | WARNING | [ ] Detected usage of meta_value, possible slow query.
 136 | ERROR   | [x] Closing PHP tag must be on a line by itself
--------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 3 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------------


FILE: /var/www/html/antispambee/wp-content/plugins/antispam-bee/antispam_bee.php
--------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------------
 269 | ERROR | [x] Each array item in a multi-line array declaration must end in a
     |       |     comma
--------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------------

Time: 4.93 secs; Memory: 36MB
```

will have a deeper look during the weekend. but looks good imo. not happy about the meta_query, but i guess no one is and no better solution in sight. since currently our acc test suite is broken, i will run those tests locally to see if sth is broken any i will try to add some tests for this functionality.

thank you @2ndkauboy

antispam_bee.php
2ndkauboy: I really dislike the idea of adding actions and fitlers using an array like this. If we want to make the plugin code more readable, we should simply remove the many unnecessary line breaks that are used in most plugins. Then every add_filter/add_action is just a single line and not 7+ lines.

derpixler: @2ndkauboy y'r right but what about the redundant `Antispam_Bee_Columns` string

2ndkauboy: It's just because all callbacks are static functions. Would you also consider it "redundant" if it would be a `$this` instead? I don't see that as an issue. Having just one giantic file with only static functions is the bigger issue for me ;)

derpixler: i think `$this` will not work. In this case `$this` is not the `Antispam_Bee_Columns` class.
But y'r right the hole file should refactored.

2ndkauboy: Yes, `$this` would not work. I was just saying that I don't see the class name as redundant.

We should probably refactor the code base for version 3.0 to make it more readable.

2ndkauboy:
> not happy about the meta_query

How would you query on meta data without a meta query? We should label this line as a false positive.

antispam_bee.php
derpixler: so then i'm quiet

websupporter: Regarding Refactoring: https://github.com/pluginkollektiv/antispam-bee/tree/version-3-0 I need to make this a PR for more visibility.

websupporter:
>> not happy about the meta_query

> How would you query on meta data without a meta query? We should label this line as a false positive.

superfluous comment by me. yes should be marked as false positive.

Btw. the tests are working again in the `master` branch :)

websupporter:
* merged `master` to bring back the tests
* fixed code style and translation functions
* made sure a spam reason only appears once in the `<select>`

websupporter:
Question I am left with:
image
Does it make sense/Can we have a select for "no reason" or "manually added"? Haven't really looked into the `\WP_Comment_Query` but I think, if we would really `$query->set('meta_query', $args)` this should be doable (`'compare' => 'NOT EXIST'`)

__
sidenote regarding the screenshot: the "no reason" comment is a viag\ra-comment before I activated the regex filter.

Zodiac1978:
> Does it make sense/Can we have a select for "no reason" or "manually added"?

We could use the `comment_unapproved_to_spam` and `comment_approved_to_spam` actions to add the "manually added" reason as comment meta.

websupporter:
My thought was more: We could just call all comments in spam folder without a reason "manually"

Zodiac1978:
> My thought was more: We could just call all comments in spam folder without a reason "manually"

But that would be not true. Comments which got marked as spam through the blacklist are there too:
https://wordpress.org/support/article/settings-discussion-screen/#comment-blacklist

websupporter:
Thanks @Zodiac1978,
let's rename it than. My goal is to be able to also filter for those spam which has no Antispam Bee reason. Maybe instead of "manual" we could say "other"?

Zodiac1978:
> My goal is to be able to also filter for those spam which has no Antispam Bee reason.

I was hoping to add more specific reasons from core if there are filter available. If this would be true we can set `manual` and `blacklist` (and if there are more than this, even more reasons).

But if all of them trigger `comment_unapproved_to_spam` or `comment_approved_to_spam` then `Other` could be a good solution.

Zodiac1978:
@websupporter Can we merge this PR now or do we have to do something?

websupporter:
:+1: lets merge it and see if what we discussed might pop up as a feature request :)

PR #307, opened 2019-11-11
Add GitHub Actions for automated deployments

2ndkauboy:
To be merged before the next release.


websupporter:
Hi @2ndkauboy,
if I understand it correctly, the README.md would be used for the readme.txt in the  `Plugin asset/readme update`. But they are not the same currently.

2ndkauboy:
You are right @websupporter. I was just trying to get this branch finished for every repository before I was kicked out of the Hackathon, that I missed this. We can simply remove the `README_NAME` env variable so the action will use the `readme.txt` again.

PR #312, opened 2019-12-08
Overhaul how we retrieve the client IP

### LINKED ISSUE #286: get_client_ip may fail if HTTP_FORWARDED doesn’t contain an IP address

websupporter:
Fixes #286
Based of #297 

In order to solve the problem in #286 we now run through all possibilities where we could get an IP. The new `_sanitize_ip()` method sanitizes what we might have got from those sources. If we haven't received a valid IP, we use the default `$_SERVER['REMOTE_ADDR']` to detect the clients IP.

Just reordering, as suggested in #286 didnt feel sufficient as the REMOTE_ADDR itself might contain a "wrong" address, e.g. the varnish address or sth else.


antispam_bee.php
```diff
@@ -2348,32 +2350,50 @@ public static function get_client_ip() {
...
-				$ip,
+			return (string) filter_var(
+				$raw_ip,
 				FILTER_VALIDATE_IP
 			);
 		}
 
-		return preg_replace(
+		return (string) preg_replace(
 			'/[^0-9a-f:\., ]/si',
```
stklcode: Comma should no longer occur at this point and escaping the dot within the character class is redundant.

websupporter: good catch

PR #321, opened 2020-03-17
Remove Fake IP check

### LINKED ISSUE #304: Fake IP check is broken
There is a check in our plugin called "Fake IP", but it is broken for IPv6, because the check compares the IP and the hostname which always returns `TRUE` - so every IPv6 comment is marked as spam.
https://github.com/pluginkollektiv/antispam-bee/blob/d7ad7fba3712ce99144bae56d08ce38807d6b5fa/antispam_bee.php#L1786

Maybe the IPv4 check is broken too, because a host can have more than one IP address and the host-to-ip/ip-to-host check does not necessarily match every time:
https://wordpress.org/support/topic/fake-ip-problem/

Should we remove the whole check? Can everyone please check if this check is working or if it only produces false positives? Thanks!

websupporter:
fixes #304 

This Pull Request removes the Fake IP check.

(Sorry for the delay, like probably everyones, my head is quite elsewhere right now.)

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1405,7 +1405,6 @@ class Antispam_Bee {
 			);
 		}
 
-
 		if ( $options['regexp_check'] && self::_is_regexp_spam(
 			array(
 				'ip'     => $ip,
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -394,7 +394,6 @@ class Antispam_Bee {
 
 		self::$defaults = array(
 			'options' => array(
-				'advanced_check'           => 1,
 				'regexp_check'             => 1,
 				'spam_ip'                  => 1,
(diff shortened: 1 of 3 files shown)


Zodiac1978:
> (Sorry for the delay, like probably everyones, my head is quite elsewhere right now.)

No need to apologize! It would have been fine to just not move this to v3 ...

But thank you very much for the PR. :)

At the moment the test is failing, but I think it is because the Behat test is still trying to test the "spam_ip" option. Correct?

websupporter:
This was due to the test suite itself again. A component of this suite had an update, which made those tests fail. I've updated this.

In the long run. It looks like wordhat (essential part of the suite) is abandoned. So in the long run we will need to switch to another stack unfortunately :(

PR #323, opened 2020-03-29
Add comment user agent to regex pattern check (fixes #322)

### LINKED ISSUE #322: Add user agent to regex check
This feature request is coming from the support forums:
https://wordpress.org/support/topic/feature-request-block-comments-depending-on-user-agent/

The information is in the database. The `wp_comments` table has a field `comment_agent` which contains the user agent of the comment author:
https://codex.wordpress.org/Database_Description#Table:_wp_comments

We could add this to this block:
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L1354-L1358

`$useragent      = self::get_key( $comment, 'comment_agent' );`

and to this block:
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L1408-L1415

`'useragent'     => $useragent,`

and to this block:
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L1526-L1530

`'useragent',`

This should be easy to add and would open up the user agent info to the regex pattern feature.

Zodiac1978:
Fixes #322

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1351,11 +1351,12 @@ class Antispam_Bee {
 	 * @return  array|void     Array with suspected reason
 	 */
 	private static function _verify_comment_request( $comment ) {
-		$ip     = self::get_key( $comment, 'comment_author_IP' );
-		$url    = self::get_key( $comment, 'comment_author_url' );
-		$body   = self::get_key( $comment, 'comment_content' );
-		$email  = self::get_key( $comment, 'comment_author_email' );
-		$author = self::get_key( $comment, 'comment_author' );
+		$ip        = self::get_key( $comment, 'comment_author_IP' );
+		$url       = self::get_key( $comment, 'comment_author_url' );
+		$body      = self::get_key( $comment, 'comment_content' );
+		$email     = self::get_key( $comment, 'comment_author_email' );
+		$author    = self::get_key( $comment, 'comment_author' );
+		$useragent = self::get_key( $comment, 'comment_agent' );
 
 		if ( empty( $body ) ) {
 			return array(
@@ -1407,12 +1408,13 @@ class Antispam_Bee {
 
 		if ( $options['regexp_check'] && self::_is_regexp_spam(
 			array(
-				'ip'     => $ip,
-				'rawurl' => $url,
-				'host'   => self::parse_url( $url, 'host' ),
-				'body'   => $body,
-				'email'  => $email,
-				'author' => $author,
+				'ip'        => $ip,
+				'rawurl'    => $url,
+				'host'      => self::parse_url( $url, 'host' ),
+				'body'      => $body,
+				'email'     => $email,
+				'author'    => $author,
+				'useragent' => $useragent,
 			)
 		) ) {
 			return array(
@@ -1528,6 +1530,7 @@ class Antispam_Bee {
 			'body',
 			'email',
 			'author',
+			'useragent',
 		);
 
 		$patterns = array(


Zodiac1978:
It looks like this doesn't work because the commentdata is coming from the hook `preprocess_comment` which us used by `handle_incoming_request` and uses the data with `_handle_spam_request`.

On `preprocess_comment` it seems that "user agent" is not filled out yet.

The function `wp_handle_comment_submission` does not contain the `comment_agent`.  This happens later in `wp_new_comment` **after** the filter hook.

I think this could be handled with just one line. Something like:
`$useragent = isset( $_SERVER['HTTP_USER_AGENT'] ) ? $_SERVER['HTTP_USER_AGENT'] : '';`

But this is just a very unreliable info and easy to spoof, so I am not very keen to get this into ASB. Especially because it could be easily solved on the server level with a few lines in the .htaccess file:
```
RewriteEngine On
RewriteCond %{HTTP_USER_AGENT} python-requests/2.23.0 [NC]
RewriteRule .* - [F,L]
```

Closing?

jaegerschnitzel:
If this doesn't work due to technical limitations you can close my request.
The .htaccess technique should fulfill my needs. Thanks for all your effort.

Zodiac1978:
And for internal documentation purpose:
WordPress can be used to block user agent strings via discussion settings. It was just not documented in the description text (will be fixed in 5.5):
https://core.trac.wordpress.org/ticket/49902

Therefore this feature isn't necessary.

Zodiac1978:
> On preprocess_comment it seems that "user agent" is not filled out yet.

This will be fixed with WordPress 5.6:
https://core.trac.wordpress.org/ticket/51044

Reopening for new discussion about including or not.

florianbrinkmann:
Does Antispam Bee mark comments that are matched by the core blocklist feature as spam, or are they moved to trash like the description says? If moved to trash, I am for adding the user agent to the regex pattern check. Otherwise, when the user could use the core feature with the same effect, I would say we do not need it.

Zodiac1978:
> Does Antispam Bee mark comments that are matched by the core blocklist feature as spam, or are they moved to trash like the description says?

@florianbrinkmann The core moderation list marks as "Pending" and the the core disallowed/blocklist moves to trash. See also #337

florianbrinkmann:
Ah, thanks for the issue link. I think it would be cool to reuse the core blocklist input. So maybe just add a checkbox to the ASB settings that lets the user change if the blocklist feature should trash or mark as spam (via the `pre_comment_approved` mentioned in #337). Having one input that is used by ASB on a totally different screen is not ideal, but then we would not reimplement something that is already available in core.

Zodiac1978:
> Having one input that is used by ASB on a totally different screen is not ideal, but then we would not reimplement something that is already available in core.

This feature would be much mightier. You could combine items (mail, body, author, email, user agent, etc.) and you could use RegEx - the core feature is just checking partials of each line, so if you try to block "bitch", you would also block "bitchange".

But the only request in the support forums for this was blocking python-requests (default user agent for Python http requests) and for this feature it would be sufficient to use the core feature.

florianbrinkmann:
> This feature would be much mightier. You could combine items (mail, body, author, email, user agent, etc.) and you could use RegEx - the core feature is just checking partials of each line, so if you try to block "bitch", you would also block "bitchange".

Oh okay, thought the core feature would also support regex. Then I would be for adding the user agent to the regex check :)

florianbrinkmann (approved):
LGTM, tested it locally and it worked like expected.

pfefferle (approved):
LGTM

PR #325, opened 2020-04-02
Add spam reason for manually marked spam

### LINKED ISSUE #324: Add spam reason for marking spam manually
There are two actions `comment_unapproved_to_spam` and `comment_approved_to_spam` we can use for this.

Maybe this action is also used if the comment is marked as spam through the blocklist in the discussion settings.

After some help from @websupporter I now  know we need to add this on more places:

https://github.com/pluginkollektiv/antispam-bee/blob/master/inc/columns.class.php#L42-L55

and 

https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L431-L441

I will prepare a branch wit this feature shortly.

Zodiac1978:
Fixes #324 

The action hooks `comment_unapproved_to_spam` and `comment_approved_to_spam` are not used for comment moderation list or comment blocklist (from discussion settings) as the first is setting as "Pending" and the last is sending to "Trash". This means the filter is just used if the spam is marked manually. I think. ;)

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -103,6 +103,22 @@ class Antispam_Bee {
 			)
 		);
 
+		add_action(
+			'comment_unapproved_to_spam',
+			array(
+				__CLASS__,
+				'update_antispam_bee_reason',
+			)
+		);
+
+		add_action(
+			'comment_approved_to_spam',
+			array(
+				__CLASS__,
+				'update_antispam_bee_reason',
+			)
+		);
+
 		if ( ( defined( 'DOING_AJAX' ) && DOING_AJAX ) || ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) ) {
 			return;
 		}
@@ -438,6 +454,7 @@ class Antispam_Bee {
 				'lang'          => esc_attr__( 'Comment Language', 'antispam-bee' ),
 				'regexp'        => esc_attr__( 'Regular Expression', 'antispam-bee' ),
 				'title_is_name' => esc_attr__( 'Identical Post title and blog title', 'antispam-bee' ),
+				'manually'      => esc_attr__( 'Manually', 'antispam-bee' ),
 			),
 		);
 	}
@@ -2390,6 +2407,21 @@ class Antispam_Bee {
 			'antispam_bee_reason'
 		);
 	}
+
+	/**
+	 * Updates the Antispam Bee reason for manual transitions
+	 *
+	 * @since   2.9.2
+	 * @param  [int] $comment Comment Object or comment ID.
+	 */
+	public static function update_antispam_bee_reason( $comment ) {
+		if ( ! is_object( $comment ) ) {
+			$comment = get_comment( $comment );
+		}
+		$id = $comment->comment_ID;
+		update_comment_meta( $id, 'antispam_bee_reason', 'manual' );
+	}
+
 
 	/**
 	 * Get the current post ID.
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -2412,7 +2412,7 @@ class Antispam_Bee {
 	 * Updates the Antispam Bee reason for manual transitions
 	 *
 	 * @since   2.9.2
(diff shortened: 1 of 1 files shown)


antispam_bee.php
```diff
@@ -2391,6 +2408,21 @@ public static function delete_spam_reason_by_comment( $comment_id ) {
 		);
 	}
 
+	/**
+	 * Updates the Antispam Bee reason for manual transitions
+	 *
+	 * @since   2.9.2
+	 * @param  [int] $comment Comment Object or comment ID.
```
websupporter, line 2415: Isn't this a WP_Comment object? 
https://github.com/WordPress/WordPress/blob/8f1eddef1176dc495ff5e9043617685adf16c62e/wp-includes/comment.php#L1730

antispam_bee.php
```diff
@@ -2391,6 +2408,21 @@ public static function delete_spam_reason_by_comment( $comment_id ) {
...
+	 */
+	public static function update_antispam_bee_reason( $comment ) {
+		if ( ! is_object( $comment ) ) {
+			$comment = get_comment( $comment );
+		}
+		$id = $comment->comment_ID;
+		update_comment_meta( $id, 'antispam_bee_reason', 'manually' );
```
websupporter, line 2422: I would just
`update_comment_meta( $comment->comment_ID, 'antispam_bee_reason', 'manually' );`

antispam_bee.php
```diff
@@ -2391,6 +2408,21 @@ public static function delete_spam_reason_by_comment( $comment_id ) {
...
 	}
 
+	/**
+	 * Updates the Antispam Bee reason for manual transitions
+	 *
+	 * @since   2.9.2
+	 * @param  [int] $comment Comment Object or comment ID.
+	 */
+	public static function update_antispam_bee_reason( $comment ) {
+		if ( ! is_object( $comment ) ) {
```
websupporter, line 2418: should we check if `get_comment_meta($comment->ID, 'antispam_bee_reason', true)` has already a reason? If someone approves a spam comment and decides this was a mistake, the original reason would remain :thinking: 

But not quite sure, if that's a good idea.

antispam_bee.php
Zodiac1978: Yes, you are right. I've copied the code from a Stackexchange function, but this function needed to cover `comment_post` additionally for example which is using a comment ID. Therefore the test if it is an object. I will remove it.

antispam_bee.php
Zodiac1978: Yes, this could be written shorter in just one line then.

antispam_bee.php
Zodiac1978: If the comment is marked as spam through any other reason and the metadata is set it can be rescued to approved/pending. If it is then marked as spam manually ASB was correct and the original reason was correct, so yes, this makes sense.

What do other think?

Zodiac1978: This does not work at the moment, because we intentionally delete the reason on unspam.
https://github.com/pluginkollektiv/antispam-bee/blob/8f3e23911a0d1ab56b94cd6dbe93c363ca827106/antispam_bee.php#L98-L104

antispam_bee.php
```diff
@@ -2412,14 +2412,10 @@ public static function delete_spam_reason_by_comment( $comment_id ) {
 	 * Updates the Antispam Bee reason for manual transitions
 	 *
 	 * @since   2.9.2
-	 * @param  [int] $comment Comment Object or comment ID.
+	 * @param  [obj] $comment Comment Object.
```
websupporter, line 2415: @param  ~[obj]~ $comment Comment Object.
@param  WP_Comment $comment Comment Object.

It would be good, if it would be called a WP_Comment. The brackets are not needed.

antispam_bee.php
websupporter: Thanks for looking into it. So we leave it as it is.

websupporter:
Thank you @Zodiac1978 :heart:

PR #335, opened 2020-04-20
Add back fake IP reason to prevent notice

### LINKED ISSUE #334: Notice: Undefined index: server
Hey,

in version 2.9.2, I got the following error message once on the edit comments page (`/wp-admin/edit-comments.php`):

```
ErrorException: Notice: Undefined index: server
#7 /wp-content/plugins/antispam-bee/inc/columns.class.php(112): Antispam_Bee_Columns::filter_columns
#6 /wp-includes/class-wp-hook.php(287): WP_Hook::apply_filters
#5 /wp-includes/class-wp-hook.php(311): WP_Hook::do_action
#4 /wp-includes/plugin.php(478): do_action
#3 /wp-admin/includes/class-wp-comments-list-table.php(411): WP_Comments_List_Table::extra_tablenav
#2 /wp-admin/includes/class-wp-list-table.php(1248): WP_List_Table::display_tablenav
#1 /wp-admin/includes/class-wp-comments-list-table.php(501): WP_Comments_List_Table::display
#0 /wp-admin/edit-comments.php(345): null
```

By accessing the page again, there is no new error event.

Zodiac1978:
If there are still some spam comments with reason "Fake IP" the plugin throws a undefined index notice.

This commit re-adds the reason to prevent this.

Fixes #334

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -449,6 +449,7 @@ class Antispam_Bee {
 				'time'          => esc_attr__( 'Comment time', 'antispam-bee' ),
 				'empty'         => esc_attr__( 'Empty Data', 'antispam-bee' ),
 				'localdb'       => esc_attr__( 'Local DB Spam', 'antispam-bee' ),
+				'server'        => esc_attr__( 'Fake IP', 'antispam-bee' ),
 				'country'       => esc_attr__( 'Country Check', 'antispam-bee' ),
 				'bbcode'        => esc_attr__( 'BBCode', 'antispam-bee' ),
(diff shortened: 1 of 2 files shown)


Zodiac1978:
If we check for the existence we don't need to add the line, or do you think we should still add it @websupporter ?

Do you have a deadline for 2.9.3?

websupporter:
I think we should still add it. Although it is unlikely, there could be stored some other data in the database under the `antispam_bee_reason` key. In this case, we should `continue;`

My idea for the deadline is June 1st as the latest :)

PR #336, opened 2020-04-22
Do not add JavaScript when in AMP endpoint

### LINKED ISSUE #283: AMP compatibility
> it injects custom scripts which not amp compatible and added a style that has “!important” which cause an error too.
> 
> Errors:
> 1.The attribute ‘style’ in tag ‘textarea’ is set to the invalid value ‘padding:0;clip:rect(1px, 1px, 1px, 1px);position:absolute !important;white-space:nowrap;height:1px;width:1px;overflow:hidden;’
> 
> 2. Custom JavaScript is not allowed
> 
> PS: when I disable the antispam bee plugin. those two errors will be resolved.
> You can validate the page above by copying the source code via page source and paste it in https://validator.ampproject.org/

From: https://wordpress.org/support/topic/not-compatible-in-amp/

Any ideas @schlessera @websupporter ?

websupporter:
This fixes #283 

We check if we are in and AMP endpoint and if so, we do not load our JavaScript .

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1222,7 +1222,9 @@ class Antispam_Bee {
 		$id_script = '';
 		if ( ! empty( $matches['id1'] ) || ! empty( $matches['id2'] ) ) {
 			$output   .= 'id="' . self::get_secret_id_for_post( self::$_current_post_id ) . '" ';
-			$id_script = '<script data-noptimize type="text/javascript">document.getElementById("comment").setAttribute( "id", "a' . substr( esc_js( md5( time() ) ), 0, 31 ) . '" );document.getElementById("' . esc_js( self::get_secret_id_for_post( self::$_current_post_id ) ) . '").setAttribute( "id", "comment" );</script>';
+			if (! self::_is_amp()) {
+				$id_script = '<script data-noptimize type="text/javascript">document.getElementById("comment").setAttribute( "id", "a' . substr(esc_js(md5(time())), 0, 31) . '" );document.getElementById("' . esc_js(self::get_secret_id_for_post(self::$_current_post_id)) . '").setAttribute( "id", "comment" );</script>';
+			}
 		}
 
 		$output .= ' name="' . esc_attr( self::get_secret_name_for_post( self::$_current_post_id ) ) . '" ';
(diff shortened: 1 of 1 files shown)


bueltge (approved):

antispam_bee.php
```diff
@@ -2193,10 +2193,14 @@ private static function _is_mobile() {
...
 	 *
+	 * Starting with v2.0, amp_is_request() is the preferred method to check,
+	 * but we fall back to the then deprecated is_amp_endpoint() as needed.
+	 *
 	 * @return bool
 	 */
 	private static function _is_amp() {
-		return function_exists('is_amp_endpoint') && is_amp_endpoint();
+		return ( function_exists( 'amp_is_request' ) && amp_is_request() )
+		       || ( function_exists('is_amp_endpoint') && is_amp_endpoint() );
```
Zodiac1978, line 2203: There some space missing. `function_exists('is_amp_endpoint')` needs to be `function_exists( 'is_amp_endpoint' )`

This is why the test is failing.

antispam_bee.php
```diff
@@ -1222,7 +1222,9 @@ public static function replace_comment_field_callback( $matches ) {
 		$id_script = '';
 		if ( ! empty( $matches['id1'] ) || ! empty( $matches['id2'] ) ) {
 			$output   .= 'id="' . self::get_secret_id_for_post( self::$_current_post_id ) . '" ';
-			$id_script = '<script data-noptimize type="text/javascript">document.getElementById("comment").setAttribute( "id", "a' . substr( esc_js( md5( time() ) ), 0, 31 ) . '" );document.getElementById("' . esc_js( self::get_secret_id_for_post( self::$_current_post_id ) ) . '").setAttribute( "id", "comment" );</script>';
+			if (! self::_is_amp()) {
```
Zodiac1978, line 1225: There are some space missing:
`if (! self::_is_amp()) {` needs to be `if ( ! self::_is_amp() ) {`, I think.

websupporter: :+1: fixed.

antispam_bee.php
websupporter: :+1: fixed.

websupporter:
Last tests fail because of
```
Then I select "Mark as Spam" from "bulk-action-selector-top"
```
"Mark has Spam" has been renamed in the latest WP versions. This has been taken care of in `master`.

Issue #73, opened 2017-01-29
Collect spam data in a smart way


schlessera:
Anonymously collect non-detected spam comments.

## What data to collect:
Comments that were not detected as spam and for which the site user manually clicked the "Spam" button.

## When to collect:
When the site user first clicks this "Spam" button, we should ask the permission to anonymously send the comment data to a centralized database, in order to improve Antispam Bee.

## How to collect:
At first, send to a HTTPS endpoint that stores everything in a simple database (probably NoSQL). We may need to evaluate a more scalable solution in the future. The collected data must not contain any mention of the sender or information about their user or system. It should contain as much information as possible about the actual spam content and where it originated.

timse201:
are we or the user allowed to do that?
are there some copyright issues?
and if we do it everytime there could be some false positives because someone dislikes an user and marks them as spam or its only because someone posted a comment several times by misklicking or caching issues etc.

but i agree
if we are allowed to (no law issues) then we should make it simpler to submit spam

websupporter:
I think, its a great idea. We should also include false positives.

I do not really see legal issues. In my understanding, if someone posts a comment, he gives the website owner the right to publish it. But honestly, I do not know how far this right can be stretched.

> there could be some false positives
Yes, but right now, we have the same issue with our Google document. I think its worth a shot.

There should be an option in the settings like (send always, never send), maybe instead but as a addition to the question "do you want to send this specific comment?" to guarantee a quicker work flow.

schlessera:
An alternative would be to add a separate button besides the `Spam` & `Trash` buttons. Something like `Send for Analysis` or similar. If they just want to get rid of their uninteresting newsletters, they will probably not click on `Send for Analysis` for these...

schlessera:
And, yes, the original idea was to ask for permission once on clicking `Spam` and then have this be the new default.

Zodiac1978:
We could use the transformation action hooks `comment_unapproved_to_spam` and `comment_approved_to_spam` or we could provide a button / action link for this.

Possible problems: Privacy concerns (IP, Mail, Content, etc. from Comments) are submitted to us (or a Third-Party-Service like Google Forms).

This feature needs consent from the user:
https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/#7-plugins-may-not-track-users-without-their-consent

krafit:
[...]

PR #344, opened 2020-08-13
Add report spam link

Zodiac1978:
Add report spam action link to spam list (#73)

### DIFF
--- a/inc/columns.class.php
+++ b/inc/columns.class.php
@@ -154,4 +154,38 @@ final class Antispam_Bee_Columns {
 		</style>
 		<?php
 	}
+
+	/**
+	 * Add comment action link to report spam to ASB
+	 *
+	 * @since 2.9.3
+	 *
+	 * @param array   $actions Array of actions.
+	 * @param comment $comment Comment object.
+	 */
+	public static function add_report_comment_action_link( $actions, $comment ) {
+
+		// URLencode comment data.
+		$name    = rawurlencode( $comment->comment_author );
+		$email   = rawurlencode( $comment->comment_author_email );
+		$ip      = rawurlencode( $comment->comment_author_IP );
+		$host    = rawurlencode( gethostbyaddr( $ip ) );
+		$url     = rawurlencode( $comment->comment_author_url );
+		$content = rawurlencode( $comment->comment_content );
+		$agent   = rawurlencode( $comment->comment_agent );
+
+		// Build action link.
+		$target = ' target="_blank" ';
+		$rel    = ' rel="noopener noreferrer" ';
+		$href   = 'href="https://docs.google.com/forms/d/e/1FAIpQLSeQlKVZZYsF1qkKz7U78B2wy_6s6I7aNSdQc-DGpjeqWx70-A/viewform?c=0&w=1&entry.437446945=' . $name . '&entry.462884433=' . $ip . '&entry.1346967038=' . $host . '&entry.121560485=' . $email . '&entry.1210529682=' . $url . '&entry.1837399577=' . $content . '&entry.372858475=' . $agent . '" ';
+
+		$action  = '';
+		$action .= "<a $target $href $rel>";
+		$action .= __( 'Report to Antispam Bee', 'antispam-bee' );
+		$action .= '</a>';
+
+		$actions['report_spam trash'] = $action;
+
+		return $actions;
+	}
 }
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -256,13 +256,21 @@ class Antispam_Bee {
 							'print_column_styles',
 						)
 					);
-
 					add_filter(
 						'manage_edit-comments_sortable_columns',
 						array(
 							'Antispam_Bee_Columns',
 							'register_sortable_columns',
 						)
+					);
+					add_filter(
+						'comment_row_actions',
+						array(
+							'Antispam_Bee_Columns',
+							'add_report_comment_action_link',
+						),
+						10,
+						2
 					);
 					add_action(
 						'pre_get_comments',
@@ -295,7 +303,6 @@ class Antispam_Bee {
 					'populate_post_id',
 				)
 			);
-
 			add_action(
 				'template_redirect',
 				array(
@@ -2796,6 +2803,7 @@ class Antispam_Bee {
 		return $current_version === self::$db_version;
 
 	}
+
 }
 
 


inc/columns.class.php
```diff
@@ -154,4 +154,38 @@ public static function print_column_styles() {
...
+	 *
+	 * @param array   $actions Array of actions.
+	 * @param comment $comment Comment object.
+	 */
+	public static function add_report_comment_action_link( $actions, $comment ) {
+
+		// URLencode comment data.
+		$name    = rawurlencode( $comment->comment_author );
+		$email   = rawurlencode( $comment->comment_author_email );
+		$ip      = rawurlencode( $comment->comment_author_IP );
```
websupporter, line 171: I am not sure about the GDPR implications, when we collect those values.

Zodiac1978: We already do this now. This is just a a helpful link to prefill the form.

But if we want to follow this path. My research says:
```
        Consent must be “freely given, specific, informed and unambiguous.”

        Requests for consent must be “clearly distinguishable from the other matters” and presented in “clear and plain language.”

        Data subjects can withdraw previously given consent whenever they want, and you have to honor their decision. You can’t simply change the legal basis of the processing to one of the other justifications.

        Children under 13 can only give consent with permission from their parent.

    You need to keep documentary evidence of consent.

The sixth legal basis is to have a “legitimate interest” to process the person’s data. 
```

I think the legitimate interest in fighting spam outweighs the rights of any spammer (or false positive user). But I am not a lawyer. I try to ping someone from the community to have look.

PraetorIM: From a legal point of view, I would have no problems with this.

In terms of **data protection**, this is possible without consent as it is within the context of a legitimate interest. (In the case of a legitimate interest, an opt-out option is required, but that is easily possible for the spammer, so you don't need to think of it.)

Also I don't see any major problems in terms of **copyright law**, because on the one hand, emails of this kind regularly do not have the necessary level of creation for copyright protection, on the other hand, the author has also consented to storage and processing by sending, because that is exactly the purpose of his mail.

So: just make it!

websupporter: Thank you @PraetorIm

websupporter:
After the discussion: This is a nice feature. Thank you @Zodiac1978 for introducing it :heart: 

Imo good to merge.

Zodiac1978:
Maybe this is important to know. This PR adds the report link just to spam links. This is because of the already existing code.

Should we add this report link to all comments?

inc/columns.class.php
```diff
@@ -154,4 +154,38 @@ public static function print_column_styles() {
 		</style>
 		<?php
 	}
+
+	/**
+	 * Add comment action link to report spam to ASB
```
schlessera, line 159:
```suggestion
	 * Add comment action link to report spam to ASB.
```

inc/columns.class.php
```diff
@@ -154,4 +154,38 @@ public static function print_column_styles() {
...
+	 *
+	 * @since 2.9.3
+	 *
+	 * @param array   $actions Array of actions.
+	 * @param comment $comment Comment object.
```
schlessera, line 164:
```suggestion
	 * @param string[]   $actions Array of actions.
	 * @param WP_Comment $comment Comment object.
```

schlessera (changes requested):

inc/columns.class.php
schlessera: Going through the discussion here, I get the impression that we are only looking at whether we as the pluginkollektiv get the user's consent to get access to this data.

However, what we are doing here is giving immediate access to a third-party (Google) via these unencrypted URL values. The way I see it, we would also need to inform the user about which third-parties we'd also give access to this data and need the user's consent for that as well.

So, even if we say that we don't need consent in this case, we still need to let the user know about the third-parties involved, no?

@PraetorIM It's not clear to me whether you consider this (we're sending unencrypted user data via a URL to a Google application), so I'd be grateful for any further clarification you can provide.

Zodiac1978:
@websupporter I don't know why Travis is still using the old behat test. Even if I restart the job it is not using the changed test and therefore fails because of the changed string.

inc/columns.class.php
PraetorIM: @schlessera I've seen that. Therefore my suggestion to go via the legitimate interest (i.e. with mention and opt-out option in the data protection declaration or at the comment-form).

websupporter:
You would need to merge the master I think. But currently master seems to fail for coding standards :smile: 

Imo its good to merge from the test side, as we know it just fails for the outdated test setup.

websupporter:
Ah sorry, I see it now I think. @schlessera did the first commit into an old branch (d53187d244ba28b7da2cb5f819576b416a8e345e) and than it was merged it into this one here (dd4ab8655bbb84a3f41a824e42be7c1bbe58375c). This branch here has the updated tests, this is why the first test fails while the second works imo.

inc/columns.class.php
Zodiac1978: > with mention and opt-out option in the data protection declaration or at the comment-form

What do you mean with that?

This PR adds a link to every spam comment in the admin area which is a link called "Report to Antispam Bee" which is going to a Google Form page and prepopulates it with the data from the comment.

It is just an external link IMHO. Everyone can see where it goes. If I need to mention it in a data protection declaration, I would need to mention every external link to a non-EU website ... or not?

And to which data protection declaration should we add it? On the website where it is used? This would not be working out I think. And why should we add it to the comment form? It is just added in the admin area and it does not do anything in the frontend. And how should I provide an opt-out for a link? You click it or you don't click it. How can we opt-out there?

I think we are all not in the same boat and speak about completely different things here.

Maybe we can talk about this in a chat. Maybe in German. To get this sorted out.

Zodiac1978: @schlessera @PraetorIM @websupporter I really would like to get 2.9.3 released soon. Any chance to get this legal question solved?

Zodiac1978:
Due to missing feedback I have punted this PR to milestone 2.10. :(

inc/columns.class.php
PraetorIM: First of all: The reporting of a URL is irrelevant in terms of data protection law, since a URL (apart from extreme exceptional cases) is not a personal date. Even a URL like firstname-lastname.de is initially without privacy problems (hard to believe in today's hysteria, I know). This would only become personal data if the person of the commentator / spammer was also transmitted. But that's not what you've planned. So just get that feature in.

Zodiac1978: @PraetorIM We are not just reporting an URL. 

We are reporting what was entered as author name, e-mail and website url. Additionally the IP address used by the visitor, the host (for this IP), the content of the comment and the user agent used.

These values are send to a Google form via URL, so that the Google form is prepopulated.

PraetorIM: Torsten, melde dich einfach einmal bei mir, ehe wir hier noch eine Brieffreundschaft anfangen ;-)

Zodiac1978:
Without clear information about the privacy implications I don't see a possible path forward here.

Additionally I will close the form and delete the excel sheet, because I am the owner of it to protect me.

If anyone is seeing a possible solution to the whole problem, please add your comment. Otherwise I don't see another  way as to close the hole issue/PR and remove any report mechanism due to unknown law implications.

florianbrinkmann:
- [ ] Add modal before sending data

Dilshadadhil (commented):
Supper

PR #355, opened 2020-10-01
Add filter to allow ajax calls

Zodiac1978:
To allow ajaxified comments, this adds a filter to allow ajax calls if necessary

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -119,7 +119,13 @@ class Antispam_Bee {
 			)
 		);
 
-		if ( ( defined( 'DOING_AJAX' ) && DOING_AJAX ) || ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) ) {
+		$disallow_ajax = apply_filters( 'antispam_bee_allow_ajax_calls', true );
+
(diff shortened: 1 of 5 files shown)


antispam_bee.php
```diff
@@ -119,7 +119,13 @@ public static function init() {
 			)
 		);
 
-		if ( ( defined( 'DOING_AJAX' ) && DOING_AJAX ) || ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) ) {
+		$disallow_ajax = apply_filters( 'antispam_bee_disallow_ajax_calls', true );
```
pfefferle, line 122: I prefer using the positive phrasing: `$allow_ajax = apply_filters( 'antispam_bee_allow_ajax_calls', false );`

It might be confusing to use `true` to disable something.

Zodiac1978: It was easier to make it this way, but I get your point.

My second thought was, "true" for default seems better than false. But I'm fine with changing the logic.

pfefferle (approved):

antispam_bee.php
pfefferle: Also valid... I approved it, because it is really a personal thing...

PR #362, opened 2020-11-01
Add WordPress version 5.4 to automated test system

### LINKED ISSUE #356: Test system needs update
WordPress version 5.4 is missing, because after WordPress 5.5 was released we just test 5.3 and then latest version and nightly version.

It seems that the old behat test is used if the branch was made before the change. At least this is the only idea I have, why the build is failing for build #739 (https://travis-ci.org/github/pluginkollektiv/antispam-bee/builds/731926635)

The PR is from today, but the branch was made for weeks/months ...

Hopefully this can be fixed, so that behat is using the corrected version.

jonathanbossenger:
As the description says, this PR adds WordPress 5.4 to the automated Travis test system

This also fixes issues with the Travis builds, in that Travis is using version 2 of composer to install composer packages, and some of the packages need updating/compatibility fixes to be included:

1. This PR updates the dealerdirect/phpcodesniffer-composer-installer package to at least version 0.7.0. 
2. This PR also adds adds the composer/package-versions-deprecated, which which provides compatibility for certain packages with Composer 2 on PHP 7.0.

It should be noted that while the builds are now running successfully, some of the tests are failing, so a good next logical step would be to fix those failing tests. My next step will be to log these as issues for future PRs.

Fixes #356


Zodiac1978:
@jonathanbossenger Many thanks for your PR! 

> My next step will be to log these as issues for future PRs.

That would be great! 

Do we need to do something to make these test available for existing PRs?

jonathanbossenger:
> Do we need to do something to make these updated tests available for existing PRs?

I honestly don't know, I don't use Travis CI much :-) I don't think it's possible however, I think one needs to re-create the PR to enable the Travis-CI builds, which runs the tests.

@schlessera might know better, I'm pretty sure he works with Travis CI more than I do, especially in WP-CLI

Now that this PR is merged I'll work on logging the test issues in the coming week, just need to update my fork first.

Zodiac1978:
> I think one needs to re-create the PR to enable the Travis-CI builds, which runs the tests.

Yes, @florianbrinkmann tried to add this on https://github.com/pluginkollektiv/antispam-bee/pull/355 but it seems that this does not work out. Re-Creating should work. I hoped for a simpler and faster solution ...

PR #367, opened 2020-11-11
Switch to GitHub Actions for tests

### LINKED ISSUE #365: Switch from Travis CI to GitHub Actions for Tests

florianbrinkmann:
I created a GitHub action workflow for the tests that are currently done by Travis. I updated the WP Behat extension (https://github.com/paulgibbs/behat-wordpress-extension) to (hopefully) get more reliable tests, but with that we loose PHP 7.0, 7.1, and 7.4 support for the Behat tests.

There is a fork of the Behat lib that seems to add support for 7.4 back (https://github.com/Liberty-Concepts/behat-wordpress-extension), the original repo is archived.

Questions:
* Is it okay for us to skip those PHP versions and only run the tests with PHP 7.2 and 7.3?
* Should we try to use the fork to get back PHP 7.4 support?

I guess we can get a better performance for the tests with caching the composer dependencies. There is an example for that in the actions repo (https://github.com/actions/cache/blob/main/examples.md#php---composer), but they use the composer.lock file there. Could we add the composer.lock file to the repo?

Closes #365


krafit:
Did we the for older PHP version with the old Behat tests?

florianbrinkmann:
Yes, the current tests run on PHP 7.0, 7.1, 7.2, 7.3, and 7.4. In the first step, I also used that implementation, but that leads to the same issues with randomly failing tests. Because of that I updated the dependencies, and that _seems_ to be better regarding those failed tests.

florianbrinkmann:
To do: check if we can fork and modify https://github.com/paulgibbs/behat-wordpress-extension to support newer PHP versions

florianbrinkmann:
I fixed the action yesterday and added PHP 7.4 and PHP 8 to the behat tests today. With that:

- Unit Tests are running for PHP 5.6, 7.2 and 7.3,
- Behat Tests for 7.2, 7.3, 7.4, 8 on WordPress versions 4.6, 4.7, 4.8, 4.9, 5.1, 5.2, 5.3, 5.4, 5.5, 5.6, nightly (not all WP versions with all PHP versions :) )
- PHPCS check on 7.2.

florianbrinkmann:
I rebased master into my branch, I am not sure if that was a good decision regarding how the commit history will look after merging :grimacing:

.github/workflows/tests.yml
```diff
@@ -0,0 +1,109 @@
...
+        composer install
+        ./vendor/bin/phpunit -c ./phpunit.xml.dist
+  test-5-4:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '5.6'
```
2ndkauboy, line 29: Ist das nun ein Test für PHP 5.4 oder 5.6? Würde auch ein Test mit PHP 5.4 funktionieren?
Is this a test for PHP 5.4 or 5.6? Would it also run with 5.4? If not, I think we can just rename it (and then also rename the `composer-php-5-4.json`?).

florianbrinkmann: Let’s change it to 5.6, and also change the `php` version number in the `composer-php-5-4.json`

pfefferle (approved):
As mentioned in the slack meeting: It looks good and functionable, so I am fine with merging it.

.github/workflows/tests.yml
florianbrinkmann: After switching it to 5.6 the test failed :D So I switched it back to 5.4, but also the PHP version of the runner. Now it works again.

florianbrinkmann:
I gave the PR a clean-up and switched the 5.4 unit tests to run on PHP 5.4 instead of 5.6. Remaining question: do we want to remove some of the old tests before merging? Some ideas:

- Remove the Unit Tests on PHP 7.2 (keeping 5.4 and 7.3)
- Only run behat tests with the last 3 WP versions + nightly version and one old but supported version, for example, 4.9? We could run the nightly and WP 5.6 on PHP 8, WP 5.5 and 5.4 on PHP 7.4 and the old WP version on PHP 7.2.

This would almost reduce the job count by half and lead to the following job list:
- test (PHP 7.3)
- test (PHP 5.4)
- behat (PHP 8, WP nightly)
- behat (PHP 8,  WP latest)
- behat (PHP 7.4, WP 5.5)
- behat (PHP 7.4, WP 5.4)
- behat (PHP 7.2, WP 4.9)
- quality (Coding-Standards)

PR #369, opened 2020-11-13
Add current screenshot for .org

### LINKED ISSUE #368: Update screenhot(s)
There was a new screenshot added in #204 but we do not added it on SVN. :(

The assets folder is a top level directory on SVN (1) , so we may need to find a way adopt these changes in our new Github actions workflow from @2ndkauboy 

Because we had no server at first, we used an own Github repo (2) to organize our assets. 

We should update those screenshots/assets and find a way to make this more easy to integrate with our actions workflow.

Maybe just add here: https://github.com/pluginkollektiv/antispam-bee/tree/master/.wordpress-org
and change the readme.txt

[1] https://developer.wordpress.org/plugins/wordpress-org/plugin-assets/

[2] https://github.com/pluginkollektiv/assets - for Antispam Bee: https://github.com/pluginkollektiv/assets/tree/master/antispam-bee/images

krafit:
Fixes #368. As mentioned over there, we are still not using the screenshot provided in #207. Since several minor changes occurred since this last screenshot, I made a new one and added it in the right place for it to be deployed to .org on the next (readme) update.


pfefferle (approved):

readme.txt
```diff
@@ -230,4 +230,7 @@ For the complete changelog, check out our [GitHub repository](https://github.com
 This update makes sure your spam check is GDPR compliant, no matter the options you choose. Please make sure to update before May 25th!
 
 ## Screenshots ##
-1. Antispam Bee settings
+1. Block or allow comments from specific countries.
+2. Allow comments only in certain languages.
+3. Add usefull spam stats to your dashboard.
```
Zodiac1978, line 235: usefull should be useful ;)

Zodiac1978:
I really like this path. I would recommend to add some more screenshots (for example the delete for specific reasons box) which is missed sometimes.

Zodiac1978 (approved):

krafit:
I'm onboard with adding more screenshots!
Let's collect some ideas, so I can produce them all at once.

pfefferle (approved):
LGTM!

PR #381, opened 2021-03-17
make the `ping` detection filterable

### LINKED ISSUE #224: Support webmention type
Webmentions aren't in WordPress Core, and exist only as a plugin. Spam webmention are not a thing yet, hopefully never. 

Webmentions are a W3C Specification and a modern successor to pingbacks. 

The only real thing Antispam Bee needs to do is be aware of the webmention comment type.

pfefferle:
For the Webmention plugin, we introduce multiple new comment types, so we need a way to hook into the `is_ping` check to return `true` on a Webmention ping.

fixes #224

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -487,7 +487,7 @@ class Antispam_Bee {
 	/**
 	 * Check if comment is a ping (pingback, trackback or something similar)
 	 *
-	 * @since   2.10
+	 * @since   2.10.0
 	 *
 	 * @param   array $comment Treated commentary data.
 	 * @return  boolean        `true` if ping and `false` if classic comment
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -487,8 +487,10 @@ class Antispam_Bee {
 	/**
 	 * Check if comment is a ping (pingback, trackback or something similar)
 	 *
-	 * @param  array $comment Treated commentary data.
-	 * @return boolean        True if ping, false if classic comment
+	 * @since   <TBD>
+	 *
+	 * @param   array $comment Treated commentary data.
+	 * @return  boolean        `true` if ping and `false` if classic comment
 	 */
 	public static function is_ping( $comment ) {
 		$types  = array( 'pingback', 'trackback', 'pings' );
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -493,7 +493,7 @@ class Antispam_Bee {
 	 * @return  boolean        `true` if ping and `false` if classic comment
 	 */
 	public static function is_ping( $comment ) {
-		$types  = array( 'pingback', 'trackback', 'pings' );
+		$types   = array( 'pingback', 'trackback', 'pings' );
 		$is_ping = false;
 
 		if ( in_array( self::get_key( $comment, 'comment_type' ), $types, true ) ) {
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -496,7 +496,7 @@ class Antispam_Bee {
 		$types  = array( 'pingback', 'trackback', 'pings' );
 		$is_ping = false;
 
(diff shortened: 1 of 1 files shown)


Zodiac1978:
Any reason why you filter the end result and not the types array?

pfefferle:
Sure. A filter to the comment type, would force new ping services to use one of the existing or a new comment type and I see this very limiting. Webmention, for example, is only the protocol and it supports several comment-types, including `comment`. If we only provide a filter for the comment types, I have to add `comment`, so the (security) mechanism to also check the request path (https://github.com/pluginkollektiv/antispam-bee/pull/381/files#diff-aed4fab3c070d2e8344bc39a125a8ddbf7ba1bda22124e940f5d9afe63e62732R1126) would not work any more or otherwise it is not possible to add `comment` webmentions via endpoint.

With the result filter, developers can also check other comment metas including the comment-type.

pfefferle:
Should I additionally add a filter for the comment types?

Zodiac1978:
> Should I additionally add a filter for the comment types?

Maybe.

I was just confused by the description:

> For the Webmention plugin, we introduce multiple new comment types

I don't know more about the logic behind Webmention and thought we just need to add more types to the array, so I thought it would be easier to just add the types to the array. If there is a Webmention type with comment which would collide with our logic it seems absolute okay to go this way.

For other uses cases it could be useful to add more comment types, but we could add this if this is an actual feature request.

florianbrinkmann:
> Any reason why you filter the end result and not the types array?

I had the same question in mind when looking at the code :D

> Sure. A filter to the comment type, would force new ping services to use one of the existing or a new comment type and I see this very limiting. Webmention, for example, is only the protocol and it supports several comment-types, including comment. If we only provide a filter for the comment types, I have to add comment, so the (security) mechanism to also check the request path (https://github.com/pluginkollektiv/antispam-bee/pull/381/files#diff-aed4fab3c070d2e8344bc39a125a8ddbf7ba1bda22124e940f5d9afe63e62732R1126) would not work any more or otherwise it is not possible to add comment webmentions via endpoint.

In my understanding, it should not make a difference whether there is one filter that filters the result, like now, or there is one filter for the types array, or am I missing something? :thinking: Plugins that use other types would have to use the filter regardless of which version is added.

Zodiac1978 (approved):
LGTM, but I recommend a second pair of eyes. (pandemic tired mode)

florianbrinkmann:
> For other uses cases it could be useful to add more comment types, but we could add this if this is an actual feature request.

Sounds like a good approacth, @pfefferle you know best what is needed for the webmention use case, so LGTM, too :+1:

florianbrinkmann (approved):

florianbrinkmann:
Because I would likely forget it: Can we already define a value for `@since`, or is that something we need to remember before the next release?

pfefferle:
@Zodiac1978 @florianbrinkmann I would also like to add the `@since`. Is there already a plan if the next release will be a major, minor or a patch?

florianbrinkmann:
I guess that a 2.10 will likely be the next release. At least I am not aware of bugs that would need a patch release. Maybe we should think about a check that we could add to the release action to search for `@since <TBD>` (or similar) in the codebase before a release is created, so we do not need to always define that when merging (or remembering it) :)

Zodiac1978:
> Is there already a plan if the next release will be a major, minor or a patch?

No, but in retrospective the last 2.9.4 should have been a minor 2.10 because of the enhancement (not just a bug fix, but adding a feature should have made it a minor ... https://semver.org/)

So, if you think this is fixing a bug, we could make it 2.9.5, if you call this adding a feature, then we should use 2.10.

Personally I am in favor of a 2.10, especially with adding the bulk analyzing feature and the decision we made in the last meeting.

Although after looking at the Milestone, I don't think we will add all those features ...

PR #384, opened 2021-03-26
Update WPCS check and introduce PHP compatibility check

### LINKED ISSUE #255: Use PHPCompatibility Check for low PHP versions
We are currently having trouble to check older PHP-versions for incompatibilities, while we have users, who still run our plugin on older PHP versions.

Underlying the PHP Compatibility Checker Plugin (https://wordpress.org/plugins/php-compatibility-checker/) is a PHP Codesniffer ruleset, which basically enables you to run static analysis on your code against specific PHP versions.

As far as I can see, we could also run PHP 5.2 checks.

Its not a 100% guarantee to not run into problems, but it would catch probably most of the common mistakes we might run into like using `array()` instead of `[]`
https://github.com/PHPCompatibility/PHPCompatibility

patrickrobrecht:
- Update the check for compliance with the WordPress Coding Guidelines to the latest version
- Fix violations with can be fixed with `phpcbf`, remove exceptions which are not necessary anymore, update ignore statements in code to new rule name (resolving the other violations is out of scope of this PR)
- Add a check for PHP compatiblity (Fixes #255)


florianbrinkmann (approved):
LGTM!

pfefferle (approved):
LGTM

PR #386, opened 2021-04-22
Add `apt-get update` before installing xvfb in GitHub Action

florianbrinkmann:
The last workflow runs for #374 failed because of 404 errors during the installation of XVFB. This PR fixes that with adding a `sudo apt-get update` before installing XVFB.


pfefferle (approved):
LGTM

Issue #391, opened 2021-05-16
Plugin incompatibility Ezoic Integration


Zodiac1978:
After installation and activation of this plugin our Honeypot is gone:
https://wordpress.org/plugins/ezoic-integration/

Not sure why this happens and if there is something we can or should do to prevent this.

First reported here: https://wordpress.org/support/topic/legitimate-comments-marked-as-spam/

Zodiac1978:
It looks like this is problem with the output buffering and maybe #357 is a related issue or a duplicate.

florianbrinkmann:
Using the `comment_form_field_comment` filter as proposed by @pfefferle seems to be the best way to overcome that issue(s), I think. I created a PR for that: https://github.com/pluginkollektiv/antispam-bee/pull/392

Issue #357, opened 2020-10-02
FacetWP compatibility


graylaurenm:
We're running into a conflict where line 1145 of antispam_bee.php uses ob_start which is never ended and interferes with other output buffers, specifically that of FacetWP in this case.

Zodiac1978:
Hi @graylaurenm  and thank your for this issue!

After looking at https://www.php.net/manual/de/function.ob-start.php and our code, I think you are right and this is a bug.

To speed things up: Is there an easy way I can reproduce the issue which you can share with us?

Zodiac1978:
There is a ticket about best practices:
https://core.trac.wordpress.org/ticket/43258

On this Github readme are some examples mentioned:
https://github.com/nextend/wp-ob-plugins-themes/blob/master/README.md

And a testing plugin:
https://wordpress.org/plugins/output-buffer-tester/

graylaurenm:
@Zodiac1978 I can provide you with a staging site. I will email you those credentials.

graylaurenm:
Sent to the address I found on your website. Subject line is `Staging site for #357` 😄

Zodiac1978:
I got the mail, but not the wordpress credentials.

Can you describe how I can see the bug in action? What steps do I need to do to see them?

graylaurenm:
Sent another email with the credentials included.

To reproduce, you'll:
1. Visit the page we set up with FacetWP
2. Expand one of the drop-downs on the right-hand side
3. Select any of those options (You'll notice that you get an error that FacetWP is unable to find the query.)
4. Next, deactivate Antispam Bee and repeat steps 1-3; the filtering works as expected

Zodiac1978:
I have logged in and saw wpDiscuz active. This is replacing the native comment form completely which makes Antispam Bee mostly useless. 

I still think you are right about not ending the `ob_start` but at the moment ASB is not doing the best because our honeypot is not used. Maybe the other checks are still working but we do net support/test with wpDiscuz and other Tools like that (e.g. Jetpack Comments).

I am wondering why this works without marking all comments as spam.

Additionally I can't reproduce the problem on the mentioned page. I set the filter and it correctly filters the recipes. Although ASB is active. But I see some JS errors in the console.

Have you already fixed this manually or why is this not reproducible for me?

graylaurenm:
[...]

PR #392, opened 2021-05-17
Replace ob_start() with `comment_form_field_comment` filter

florianbrinkmann:
Some users experience issues with ASB in combination with other plugins because we do not end the `ob_start()`, or, because the other plugins end it (#391 and #357). This PR removes the `ob_start()` and uses the `comment_form_field_comment` filter instead, as proposed by @pfefferle https://github.com/pluginkollektiv/antispam-bee/issues/357#issuecomment-729597129


florianbrinkmann:
The filter should only run when the comment form is displayed, so I guess it is save to remove the checks from the function that were necessary before (?)

Zodiac1978:
Thank you for the PR!

Just to be sure: This would make the option obsolete to look for comment forms on archive pages, or not? Because the filter would fire for each comment form. Then we could remove this option too - not sure if there other implications if we do this.

florianbrinkmann:
Hm, yes, I guess you are right, that option should not be needed anymore then. I will give it a try locally and add the form to the archive page.

florianbrinkmann:
Seems to work fine. The honeypot field is also displayed in the form on the comment form.

florianbrinkmann:
We are using the value of the `always_allowed` option for two filters (https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L2733 and https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L2766). When we remove the option, we need to decide what to do with that parameter. Just removing it could lead to errors with users that use the filter with all three parameters.

pfefferle:
The question is: should antispam bee also work with contact forms or other forms in general or only with "comments". That is why I have not made a PR up to now.

Zodiac1978:
> should antispam bee also work with contact forms or other forms in general

Wouldn't those contact forms need other implementations (other Regex) and hopefully have their own filter hooks?

florianbrinkmann:
> After the change will "only" work for comment forms, that uses the comment_form_field_comment filter.

The only situation where this would not work would be when the comment form is built manually, wouldn’t it? The filter is fired in the `comment_form()` function which should be **the** function when one wants to display a comment form(?)

> […] and hopefully have their own filter hooks?

I would also think that. And if, some day, ASB can be extended, the plugin that adds support for other forms would need to handle that :)

Zodiac1978:
> After the change will "only" work for comment forms, that uses the comment_form_field_comment filter.

Is there a way to circumvent this filter, so that this approach does not work in every case?

pfefferle:
Don't get me wrong, I second the PR, I only wanted to point out that it does not work exactly like it works now.

But @florianbrinkmann is right, maybe we should remove the `name="comment"` so that it can be implemented by other forms that have a textarea. Or that we use the `<from>` as anchor instead of the `<textarea>`.

pfefferle:
...maybe not a good idea to try to make the implementation generic, but to add a generic version that can be implemented by plugin providers instead.

pfefferle (approved):

florianbrinkmann:
> Don't get me wrong, I second the PR, I only wanted to point out that it does not work exactly like it works now.

You are right, there could be cases where it works now but does not work after the change (manually crafted comment forms). Should we target that change for a major release, because it is a (theoretically) breaking change?

pfefferle:
Good point! I would second the major release version!

florianbrinkmann:
- [ ] Set option like `asb_use_output_buffer` to `0` in `register_activation_hook()` and use the buffer only when that option is not there or `1`.
- [ ] Add Checkbox to opt in for the output buffer usage.

florianbrinkmann:
I added the option to use output buffering instead of the filter. When the plugin is activated, this option is set to `0`, so the filter is used.

When the option is not set (for installations that update to the new version), the output buffering will be used and the option is checked. **Important:** to make that work, I needed to modify the `get_key()` method, so that it does not return `null` for both cases: the option is not there or the option value is `0`. That should not have effect on other code, because, for example, `! self::get_option( 'delete_data_on_uninstall' )` will be true for `self::get_option( 'delete_data_on_uninstall' )` being `0` or `null`.

florianbrinkmann:
- [ ] Replace anonymous function (because PHP 5.2…)

pfefferle (approved):
The filter seems to work and the setting is default on, if I had a previous version installed and it is not active on new installations. Works as expected!

pfefferle (approved):

PR #393, opened 2021-05-27
introduce SonarCloud analysis

stklcode:
This PR adds a _.sonarcloud.properties_ file to add some configuration for the SonarCloud [automatic analysis](https://sonarcloud.io/documentation/analysis/automatic-analysis/).

This mode does not support test coverage (obviously, without running tests), but tells the analyzer which files to scan and thus reduces a number of false positives on test files and external resources.

We might go one step further and replace the automatic by manual analysis from CI, if we do want coverage analysis. Requires some tweaks to properly work with multiple test frameworks involved though, so this seems to be a good choice to start for now.

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -12,6 +12,7 @@
 /.eslintrc.json
 /.gitattributes
 /.gitignore
+/.sonarcloud.properties
 /.stylelintrc.json
 /composer.json
 /composer.lock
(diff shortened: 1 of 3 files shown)


florianbrinkmann (approved):
LGTM

pfefferle (approved):
LGTM

PR #396, opened 2021-06-10
Change API endpoint to use iplocate.io instead

### LINKED ISSUE #388: ip2country API
There is a notice (maybe for longer, I just saw it now) on http://ip2country.info/

> Exciting News!
>
> IP2country is joining ipapi.com to offer an even more advanced IP API, now offering IP address geolocation & reverse IP lookups in real-time at an unparalleled level of accuracy. To keep using our API, switch to ipapi.com and get your free API key. 

Does anyone have more information about this? Looks like we need to remove this or find a way to use the API key without making it public.

pfefferle:
fix #388


Zodiac1978:
Nice work! 

Maybe we could add a filter to customize the URL (to support an API key `https://www.iplocate.io/api/lookup/ip/format?apikey=&callback=`) and the JSON key?

pfefferle:
@Zodiac1978 the API key filter is a good idea but the `callback` is only intresting for AJAX calls, so I see no need to add that.

> Edit: And to change to a completely different service which also uses JSON.

I like the idea in general, but then we have to also make the JSON parsing extensible, because other JSON APIs might have different responses with different keys, and that would make it very complex.

2ndkauboy:
Then we either add a filter to 1770 as well or we make the whole function filterable.

Zodiac1978:
The callback part was just a copy&paste issue. The important part is to support adding an API key if someone needs to do that.

If we make just the JSON key also filterable we would at least support ip2country.info *and* iplocate.io - if other services need different solutions we don't need to support them. That was my thought ...

2ndkauboy:
There are multiple things that we may need to make filterable. In essence the functions does this:

1. Get black and white entries (we should really change those variable names!)
2. Call the API
3. Get the 2-chars country codes from the response
4. Compare the code to the entries

How about moving 2 and 3 into a separate function and allow developers to call their own function instead. We would pass the (anonymized) IP address and would expect a 2-chars country code as a return.

pfefferle:
@2ndkauboy sounds good, but (to keep the PR simple) I would recommend to add the API key to this PR and make a new one for the restructuring.

Is that ok for you @2ndkauboy @Zodiac1978 ?

2ndkauboy:
Doesn't the new API not need a API key?

Zodiac1978:
> Doesn't the new API not need a API key?

That's true, but it is rate limited to 1000 requests per day (from one IP, I think). So if a power user *needs* to add an API key, we should make this possible.

pfefferle:
@2ndkauboy @Zodiac1978 what do you think about https://github.com/pluginkollektiv/antispam-bee/pull/396/commits/a8b204a48124b2124d2a2465c5dfbea1f02a8ebe to implement a custom IP check?

With that change we do not have to make the following customizable:

* Get the 2-chars country codes from the response
* Compare the code to the entries

I would keep it that way and check the support requests, before we implement a more "friendly" solution.

pfefferle:
Todo:

* [ ] Tests

Zodiac1978:
I needed some time to understand what you have done in the last commit. If I understand correctly this does nothing at the moment, but can be used to actually add a callback function which gets the ip and blacklisted and whitelisted countries as arguments. Correct? 

This would definitely need documentation how to use it. If I want to add my own ip check, I need to provide the check myself and use the new filter. This would be a much broader solution than just provide a filter for defining the JSON key.

And for our own new default there is now a filter to add the API key for the power users.

LGTM.

pfefferle:
> I needed some time to understand what you have done in the last commit. If I understand correctly this does nothing at the moment, but can be used to actually add a callback function which gets the ip and blacklisted and whitelisted countries as arguments. Correct?

Exactly!

> This would definitely need documentation how to use it. If I want to add my own ip check, I need to provide the check myself and use the new filter.

Good point. Will add some more documentation.

> This would be a much broader solution than just provide a filter for defining the JSON key.

I also added a filter for the API key! https://github.com/pluginkollektiv/antispam-bee/pull/396/commits/37cad2614779c9d0931a0f7b18eef57f97e9a981

Zodiac1978:
> Will add some more documentation.

I was thinking more about https://antispambee.pluginkollektiv.org/documentation/ ;)

Maybe an example implementation.

> I also added a filter for the API key!

Yes, this was just a summary for me. I think this a good 3-way solution: 
Default with free tier, API key for those who are in need of more and the new filter for those who want to implenent a complete other service.

Could this smart solution be a way to implement other services with other arguments, like if someone wants to re-add https://www.stopforumspam.com/usage for example?

2ndkauboy (approved):

florianbrinkmann:
The API is saying I am in Luxembourg, but apart from that, it works like expected :)

florianbrinkmann (approved):
LGTM. One last thing, maybe change `2.10` to `2.10.0` in the `@since` and `@change` tags.

Issue #33, opened 2016-07-12
Road to 2.9.0


swissspidy:
I moved a bunch of issues to the 2.7.0 milestone today to give us a little bit of a challenge.

I also updated master to include changes for 2.6.9. That release basically only needs to be committed to the plugin directory.

2.7.0 includes mainly bug fixes, but also some suggested features (not set in stone!). Before we implement any of these, I really want to have a solid structure so we can add proper unit tests along the way. See pluginkollektiv/antivirus#13

We should also think about a proper workflow to do stuff like:
- ~Minify CSS and JS upon build.  
  The GitHub repository shouldn't include minified assets.~
- ~Add PHP_CodeSniffer rules.
  To ensure we follow the WP coding standards.~
- Add other tools to catch mistakes like missing text domains.
- ~Use a proper deploy script.
  Committing to SVN by hand is annoying. https://github.com/stephenharris/grunt-wp-deploy/ is the de-facto standard IMHO and a new version of that script is coming out soon.~

timse201:
you can add the project to https://travis-ci.org/ which could help testing

swissspidy:
Travis is implicitly included in that list for all our plugins 🙂

timse201:
ok xD
didnt read that

websupporter:
Coding standards and travis ci for these standards have been introduced with #188

PR #401, opened 2021-06-24
Adding checks for global prefixes and text domain values

2ndkauboy:
See #33

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -97,3 +97,4 @@ jobs:
         composer install
         npm install
         composer lint-all
+        composer cs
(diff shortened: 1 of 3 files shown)


2ndkauboy:
It seems that we are not running tests for `phpcs` right now, correct? At least this is only running `eslint` and `stylelint`: https://github.com/pluginkollektiv/antispam-bee/blob/master/.github/workflows/tests.yml#L99

pfefferle (approved):
LGTM!

florianbrinkmann:
> It seems that we are not running tests for phpcs right now, correct? 

Yes, I think you are right

florianbrinkmann (approved):
LGTM

Issue #378, opened 2021-03-03
Adjust colors with WP 5.7 color palette


Zodiac1978:
New palette:
https://codepen.io/ryelle/full/WNGVEjw

Our colors:
![colors](https://user-images.githubusercontent.com/870475/109808482-76955900-7c27-11eb-88a8-d2789a7cdcfe.png)

Although there is one more, because the #0073aa is overlayed with a 0.2 opacity white, resulting in #C2DEEA.

Zodiac1978:
These colors come from the CSS files:
https://github.com/pluginkollektiv/antispam-bee/tree/master/css

and from the dashboard JS file:
https://github.com/pluginkollektiv/antispam-bee/blob/master/js/dashboard.js

Zodiac1978:
Not sure why this unprefixed `browsehappy` class is there:
https://github.com/pluginkollektiv/antispam-bee/blob/master/css/styles.css#L19-L24

According to the comment it is just used for IE lower than 9:
https://github.com/pluginkollektiv/antispam-bee/blob/5174f9ad3e0a942c9ff0f66f8fa2f583f246f048/inc/gui.class.php#L182-L186

As WP dropped IE 8,9 and 10 support, I think we can remove this section:
https://make.wordpress.org/core/2017/04/23/target-browser-coverage/

As MS dropped support for conditional comments with IE 10 we need another way to trigger this warning or we justremove it completely (this would be my recommendation).

PR #404, opened 2021-07-08
Adjust colors to fit 5.7. color palette

krafit:
see #378

### DIFF
--- a/css/dashboard.css
+++ b/css/dashboard.css
@@ -1,5 +1,5 @@
 #ab_chart {
-	color: #a0a5aa;
+	color: #a7aaad;
 	height: 140px;
 	margin: 0 -4px;
 	text-align: center;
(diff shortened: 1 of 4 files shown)


Zodiac1978:
Do you think this is a visible change? Or too subtle?
Do you have before/after screenshots?

krafit:
@Zodiac1978  It's a slight change. You can see the difference if you see old and new side by side, but I don't think it's noticeable otherwise.

Bildschirmfoto 2021-07-08 um 10 32 53
before (left), after (right)

Zodiac1978:
This looks fine and complete.

The only thing we could think about is "merging" some colors, like Grey20 and Grey30 and just go with Grey30 for example.

Or aim a higher goal and try to go even more with the new higher contrast colors. Statify seems to have this already:

Bildschirmfoto 2021-07-08 um 10 54 02

Bildschirmfoto 2021-07-08 um 10 54 23

Zodiac1978:
UI changes are handled in https://github.com/pluginkollektiv/antispam-bee/issues/406

Zodiac1978 (approved):
LGTM

PR #405, opened 2021-07-08
Change wording for country options

### LINKED ISSUE #400: Change wording for IP address lists
The plugin is currently using the terms black/white (list) for it's lists of IP addresses to be blocked or allowed. I would ask to rename them:

 - `$black` => `$blocked` or `$denied`
 - `$white` => `$allowed`
 
 There are also two options `country_black` and `country_white` which we would have to migrate to new names.

florianbrinkmann:
Fixes #400 

This PR changes the wording for the country options to `country_allowed` and `country_denied`. The migration runs:

- when the plugin is updated normally via the upgrade mechanism (I think we cannot test that before the release, sadly…)
- when the plugin is updated via ZIP upload
- when the new version of the plugin is installed via FTP or so, the migration runs when the ASB options page is visited


florianbrinkmann:
What needs to be tested (with one of the ways described in the PR description above): 

- [ ] is the `antispambee_db_version` correctly upgraded to `1.02`
- [ ] are the option names `country_white` and `country_black` changed to `country_allowed` and `country_denied` in the db and the values correctly migrated

pfefferle (approved):
LGTM

florianbrinkmann:
- [ ] Update `@change` version numbers to `2.10.0`

antispam_bee.php
```diff
@@ -93,6 +93,51 @@ class Antispam_Bee {
 	 * @change  2.6.4
```
pfefferle, line 93: update `@change`

PR #408, opened 2021-07-15
Create dependabot.yml

pfefferle:
> To get started with Dependabot version updates, you'll need to specify which package ecosystems to update and where the package manifests are located.
> Please see the documentation for all configuration options:
>
> https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "composer" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "npm" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"


pfefferle:
@stklcode what is your experience with dependabot so far?

florianbrinkmann:
> Required. You add one package-ecosystem element for each package manager that you want Dependabot to monitor for new versions. The repository must also contain a dependency manifest or lock file for each of these package managers.

If I understand that correctly, we would need a lock file or dependency manifest per package manager in the repo. We have the `composer.lock` file, but nothing for NPM.

And, one other thing: I don’t know if that is a problem because we have the `composer.lock` file, but updating Composer things crashed the GitHub action recently: https://github.com/pluginkollektiv/antispam-bee/runs/2792012508?check_suite_focus=true

pfefferle:
> dependency manifest or lock

I tested it on several repos and it also works with composer.json and package.json.

I also checked if it is possible to ignore certain dependencies like phpunit and it is possible.

stklcode:
The config so far is more or less default, but defines the scans explicitly.

There is one difference between `composer.json` and `composer.lock` (or equivalent NPM). If no lock file is present and the original declaration uses soft references `^1.2.3`, `~1` or whatever, there are no warnings about minor patches - why should they, it's updated during every installation.

If there are only dev-dependencies one could consider removing the locks. This _might_ yield instabilities in the build process, if dependencies do not follow semver perfectly or - we all know it happens - a minor patch update breaks stuff accidentially, so there are pros and cons for both ways.

Compromise could be auto-merge, i.e. we declare automerge for e.g. `semver:minor` or `semver:patch` updates, s.t. we don't have to care about a pile of open PRs of which most are approved and merged in seconds.

florianbrinkmann (approved):

PR #416, opened 2021-07-29
DocBlock rework

florianbrinkmann:
Adds the `@change` tags back that were replaced by `2.10.0` and replaces `@change` with `@since`.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -410,7 +410,6 @@ class Antispam_Bee {
 	 * Action deleting the plugin
 	 *
 	 * @since  2.4
-	 * @since  2.4
 	 */
 	public static function uninstall() {
 		if ( ! self::get_option( 'delete_data_on_uninstall' ) ) {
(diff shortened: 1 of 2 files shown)


antispam_bee.php
```diff
@@ -407,8 +409,8 @@ public static function deactivate() {
 	/**
 	 * Action deleting the plugin
 	 *
-	 * @since   2.4
-	 * @change  2.4
+	 * @since  2.4
+	 * @since  2.4
```
pfefferle, line 413: I think one `@since 2.4` is enough ;)

antispam_bee.php
```diff
@@ -583,8 +586,8 @@ public static function load_plugin_lang() {
 	/**
 	 * Add the link to the settings
 	 *
-	 * @since   1.1
-	 * @change  1.1
+	 * @since  1.1
+	 * @since  1.1
```
pfefferle, line 590: duplicate `@since`

antispam_bee.php
florianbrinkmann: thanks! :)

pfefferle (approved):
LGTM

PR #424, opened 2021-12-15
Add FAQ entry about AJAX comment plugins

### LINKED ISSUE #390: Optimize FAQ entry about AJAX comment plugins
In our FAQ we write:
> It also won’t work with any AJAX-powered comment forms.

After the merge of #355 we should tackle #376 and also we should update the FAQ to reflect this change.

Zodiac1978:
Fixes #390


krafit (approved):
LGTM thank you

PR #430, opened 2022-01-27
PHPCS Action

pfefferle:
Run PHPCS checks on every push.


pfefferle:
This is already part of the test.yml.

PR #437, opened 2022-02-04
Add aria label to honeypot textarea

### LINKED ISSUE #434: WAVE test failing
First reported here: https://wordpress.org/support/topic/comment-form-accessibility-compliance/

@Zodiac1978 
The solution is to add `aria-label="whatever"` as attribute to the textarea.

The automated tool checks if a form control has an accessible name. In this case, even if the form control is hidden from sight and screen reader, it must have an accessible name to pass the WAVE test.
This can be easily solved by adding an aria-label like `aria-label="comment" `

```
<textarea aria-label="comment" id="two" aria-hidden="true" name="comment" autocomplete="new-password" style="padding:0 !important;clip:rect(1px, 1px, 1px, 1px) !important;position:absolute !important;white-space:nowrap !important;height:1px !important;width:1px !important;overflow:hidden !important;" tabindex="-1"></textarea>
```

The aria-label can be anything, because it is never announced or visible.
And then it passes WAVE :-)

I made a [codepen with both options](https://cdpn.io/pen/debug/yLPYoJM), with and without adding aria-label. The the [first one fails, the second one passes in WAVE](https://wave.webaim.org/report#/https://cdpn.io/pen/debug/yLPYoJM).

_Originally posted by @rianrietveld in https://github.com/pluginkollektiv/antispam-bee/issues/292#issuecomment-1024962090_

Zodiac1978:
Adds an aria label to honeypot textarea to fix the failing WAVE test.

Fixes #434

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1278,7 +1278,7 @@ class Antispam_Bee {
 		$output .= $matches['between1'] . $matches['between2'] . $matches['between3'];
 		$output .= $matches['after'] . '>';
 		$output .= $matches['content'];
-		$output .= '</textarea><textarea id="comment" aria-hidden="true" name="comment" autocomplete="new-password" style="padding:0 !important;clip:rect(1px, 1px, 1px, 1px) !important;position:absolute !important;white-space:nowrap !important;height:1px !important;width:1px !important;overflow:hidden !important;" tabindex="-1"></textarea>';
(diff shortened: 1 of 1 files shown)


florianbrinkmann (approved):
LGTM

pfefferle (approved):
LGTM

PR #440, opened 2022-02-15
Raise priority for comment field filter

### LINKED ISSUE #438: Problem with theme that alter the comment form textarea
First reported here: https://wordpress.org/support/topic/howto-configure-the-honeypot/

Themes can use the same filter like we do.

For example the the Bravada (https://wordpress.org/themes/bravada/) which is using `comment_form_field_comment` to change the comment form textarea:

```
/* Edit comments form textarea: removed label and replaced it with a placeholder */
function bravada_comments_form_textarea( $arg ) {
	$arg = '<p class="comment-form-comment"><label for="comment">' . _x( 'Comment', 'noun', 'bravada' ) .
			'</label><em><textarea placeholder="'. esc_attr_x( 'Comment', 'noun', 'bravada' ) .'" id="comment" name="comment" cols="45" rows="8" aria-required="true">' .
			'</textarea></em></p>';

	return $arg;
} // bravada_comments_form_textarea()
```

Now our filter does not work anymore.

Enabling `[_] Check complete site markup for comment forms` in ASB option fixes the problem.

Using a higher priority than 10 (eg. "12" - for later execution) in ASB also fixes the problem.

There are 93 themes in the official theme directory using this filter:
https://wpdirectory.net/search/01FVDFVYDQGM702RXZ0G17YAGY

Zodiac1978:
Raise the priority for the `comment_form_field_comment` filter from default 10 to 99. This should fix most problems with themes using this filter too, but still is usable for customizing our output.

Fixes #438

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -2250,7 +2250,7 @@ class Antispam_Bee {
 		if ( function_exists( 'filter_var' ) ) {
 			return filter_var( $ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4 ) !== false;
 		} else {
-			return preg_match( '/^\d{1,3}(\.\d{1,3}){3,3}$/', $ip );
+			return preg_match( '/^\d{1,3}(\.\d{1,3}){3}$/', $ip );
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


florianbrinkmann (approved):
LGTM

pfefferle (approved):
LGTM

PR #444, opened 2022-03-20
WIP: First work on creating the rule classes

angelocali:


inc/Verifiable.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace AntispamBee\Rules;
+
+interface Verifiable {
+	public static function init();
```
2ndkauboy, line 6: Should this be part of the interface, or should we allow developers to initialize their filter differently? Would they be able to use our trait as well inside their own classes?

PR #446, opened 2022-03-20
WIP: CloudFest Hackathon - More work on Rules and Post Processors ...

angelocali:


inc/Admin/Fields/Checkbox.php
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Checkbox field.
+ */
+class Checkbox extends Field implements RenderElement {
+    /**
+     * Get HTML.
+     */
+    public function render() {
+        echo '<input type="checkbox" name="' . $this->get_name() . '" value="1" ' . checked( 1, $this->get_value(), true ) . ' />';
```
2ndkauboy: I would like us to use `printf` instead of string concats to make the code more reable:

```php
printf(
	'<input type="checkbox" name="%s" value="1" %s />',
	$this->get_name(),
	checked( 1, $this->get_value(), false )
);
```

PR #449, opened 2022-03-26
remove superfluous type attribute from inline script tag (#448)

### LINKED ISSUE #448: Warning: The type attribute is unnecessary for JavaScript resources.
Hey peeps

I'm getting this warning on an otherwise flawless HTML validation if AntiSpam Bee is activated.

image

The warning is triggered by this code:
`<script data-noptimize type="text/javascript">document.getElementById("comment").setAttribute( "id", "ae322e2e655f1c4f69db80f63425cc99" );document.getElementById("i2381a6a1a").setAttribute( "id", "comment" );</script>↩`

image

The **type="text/javascript"** is no longer needed, and has not been for a long time. Not even WP core outputs the type attribute anymore.

I've got an "HTML clean-up" snippet that filters it out for plugins that still outputs the type attribute, for a nice green HTML validation. It's a workaround, and just proves that AntiSpam Bee works fine without the type attribute. But I'd obviously prefer not to add additional filtering to the output.

Would you consider removing the type attribute?

Bjarne

stklcode:
The "type" attribute is technically not invalid, but not necessary
either, and it might raise warnings in markup validations.
Remove the attribute and add some line-breaks for readability.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1272,7 +1272,10 @@ class Antispam_Bee {
 		if ( ! empty( $matches['id1'] ) || ! empty( $matches['id2'] ) ) {
 			$output .= 'id="' . self::get_secret_id_for_post( self::$_current_post_id ) . '" ';
 			if ( ! self::_is_amp() ) {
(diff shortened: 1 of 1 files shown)


pfefferle (approved):
LGTM

Issue #458, opened 2022-06-02
Prepare new release // check compatibilty with WP 6.0


Zodiac1978:
- [x] Check WP 6.0 compatibility
- [x] Is 2.11.1 okay or has anyone something for 2.12?
- [x] Add new entry in CHANGELOG.md
- [x] Raise version number to 2.11.1/2.12 in antispam_bee.php
- [x] Raise "Tested up to:" to 6.0 in readme.txt
- [x] Raise "Stable tag:" to 2.11.1/2.12 in readme.txt
- [x] Add new entry to changelog area in readme.txt
- [x] Check for wrong or placeholder @since docblock entries
- [ ] Post a release article on https://antispambee.pluginkollektiv.org/
- [ ] Post a tweet with pluginkollektiv and/or antispambee account

PR #460, opened 2022-06-22
Prepare 2.11.1 release

florianbrinkmann:


pfefferle (approved):

Zodiac1978:
I have linked the PR with the issue (#458 ) and checked everything what is done. LGTM!

Why do you have removed the POT file? Looks like this still useful:
https://developer.wordpress.org/plugins/internationalization/localization/#pot-portable-object-template-files

Maybe this needs an update though?

florianbrinkmann:
> Why do you have removed the POT file? Looks like this still useful.

I think I created that file during a test locally and forgot to remove it again before pushing the first commit, it is not part of the `master` branch.

florianbrinkmann:
> I have linked the PR with the issue (https://github.com/pluginkollektiv/antispam-bee/issues/458 ) and checked everything what is done. LGTM!

thanks!

Zodiac1978:
> Otherwise you don't need that file

For starting a new locale this would still be helpful, or not? Loco translate recommends using a generated POT file instead of extracting the strings yourself.
https://localise.biz/wordpress/plugin/manual/templates
This would help to get a new locale up and running without requiring 90% of translations.

> I think I created that file during a test locally and forgot to remove it again before pushing the first commit, it is not part of the master branch.

This explains why I did not found the `lang` folder in `master` branch 😂.

Then I'm fine with going on ...

PR #463, opened 2022-07-28
Change link to ISO.org

### LINKED ISSUE #462: Link to ISO codes not helpful
The link for the explanation of ISO Codes despite refering to the ISO.org website is not really helpful as I brings you to a general overview where one has to identify the correct link for the list. It would be better to link directly to this list https://www.iso.org/obp/ui/#search or even make use of the wikipedia-entry https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes

Bildschirmfoto 2022-07-28 um 10 37 39

stkjj:
issue #462

### DIFF
--- a/inc/gui.class.php
+++ b/inc/gui.class.php
@@ -286,7 +286,7 @@ class Antispam_Bee_GUI extends Antispam_Bee {
 									$iso_codes_link = sprintf(
 										'<a href="%s" target="_blank" rel="noopener noreferrer">',
 										esc_url(
-											__( 'https://www.iso.org/iso/country_names_and_code_elements', 'antispam-bee' ),
+											__( 'https://www.iso.org/obp/ui/#search', 'antispam-bee' ),
 											'https'
 										)
 									);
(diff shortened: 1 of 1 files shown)


inc/gui.class.php
```diff
@@ -286,7 +286,7 @@ public static function options_page() { ?>
 									$iso_codes_link = sprintf(
 										'<a href="%s" target="_blank" rel="noopener noreferrer">',
 										esc_url(
-											__( 'https://www.iso.org/iso/country_names_and_code_elements', 'antispam-bee' ),
+											__( 'https://www.iso.org/obp/ui/#search', 'antispam-bee' ),
```
stklcode, line 289: The proposed link leads to the empty search page when opened outside an existing session and is not really helpful in this case.

Guess the intention was https://www.iso.org/obp/ui/#search/code/
(also see https://github.com/pluginkollektiv/antispam-bee/issues/462#issuecomment-1296804571 for discussion)

stklcode (changes requested):

stklcode (approved):

florianbrinkmann (approved):

PR #471, opened 2022-10-01
fix: unnecessary translation

### LINKED ISSUE #461: Unnecessary translation
**Describe the bug**
We have links to the support forum added as translatable, but although the plugin page itself is made for i18n, the support forum seems to be not.

Relevant code is here:
https://github.com/pluginkollektiv/antispam-bee/blob/4f1d2bf69ada8895f33da838c290aeb42902888e/inc/gui.class.php#L537

1. Go to https://wordpress.org/support/plugin/antispam-bee - works
2. Go to https://de.wordpress.org/support/plugin/antispam-bee/ - don't work (404)

**Expected behavior**
Obviously the support forum is not made for i10n/i18n.

First reported here: https://wordpress.org/support/topic/invalid-support-link/

2ndkauboy:
Closes #461

### DIFF
--- a/inc/gui.class.php
+++ b/inc/gui.class.php
@@ -528,7 +528,7 @@ class Antispam_Bee_GUI extends Antispam_Bee {
 							<a href="https://www.paypal.com/cgi-bin/webscr?cmd=_donations&business=TD4AMD2D8EMZW" target="_blank" rel="noopener noreferrer"><?php esc_html_e( 'Donate', 'antispam-bee' ); ?></a>
 						</p>
 						<p>
-							<a href="<?php echo esc_url( __( 'https://wordpress.org/plugins/antispam-bee/faq/', 'antispam-bee' ) ); ?>" target="_blank" rel="noopener noreferrer"><?php esc_html_e( 'FAQ', 'antispam-bee' ); ?></a>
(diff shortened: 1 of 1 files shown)


inc/gui.class.php
```diff
@@ -528,13 +528,13 @@ public static function options_page() { ?>
 							<a href="https://www.paypal.com/cgi-bin/webscr?cmd=_donations&business=TD4AMD2D8EMZW" target="_blank" rel="noopener noreferrer"><?php esc_html_e( 'Donate', 'antispam-bee' ); ?></a>
 						</p>
 						<p>
-							<a href="<?php echo esc_url( __( 'https://wordpress.org/plugins/antispam-bee/faq/', 'antispam-bee' ) ); ?>" target="_blank" rel="noopener noreferrer"><?php esc_html_e( 'FAQ', 'antispam-bee' ); ?></a>
+							<a href="<?php echo esc_url( __( 'https://wordpress.org/plugins/antispam-bee/#faq', 'antispam-bee' ) ); ?>" target="_blank" rel="noopener noreferrer"><?php esc_html_e( 'FAQ', 'antispam-bee' ); ?></a>
```
nielslange, line 531: Wouldn't it be an option to hardcode this link as well, @2ndkauboy? Similar to what you did with the other link below? I think it's very unlikely that the FAQ link will change in the near future.

websupporter: This link is actually translated:
image

But there is no supported wp.org support forum for the plugin in another language. I assume this is why the translate function has been removed.

nielslange: Ah, I see. Thanks for clarifying this, @websupporter.

Zodiac1978 (approved):
LGTM!

PR #478, opened 2022-11-01
Adapt translation loading

### LINKED ISSUE #423: Problem with translation
**Describe the bug**
We are using translation function before we are loading the translation.

**To Reproduce**
First occurrence of i18n function is here:
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L486

Localization is loaded later here:
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L578-L582

**Expected behavior**
We should load the translation file before we use any i18n functions. Otherwise we produce problems (for example i18n files are loaded twice or user locale vs. site locale is not working)

**Additional context**
Query Monitor is showing `_load_textdomain_just_in_time()` instead of the correct `load_plugin_textdomain` call in the language tab.

florianbrinkmann:
In https://github.com/pluginkollektiv/antispam-bee/issues/423#issuecomment-997199403 I proposed to just move the `load_plugin_lang` call above the call for setting up the internal vars. If we would do that, we could not check for the pages but would need to load the langs in all cases, because otherwise the loading would be over the `_load_textdomain_just_in_time` function for the views that are not inside the check (because the gettext functions in `_init_internal_vars` are executed always).

So I moved the reasons array to another method that is only added to the `$options` property, if one of the pages is displayed where lang strings are displayed.

Fixes #423


stklcode (approved):
Changes look reasonable.

Quickly clicked through several scenarios with different combinations of site and user language. Experienced no UI issues and no i18n breakpoint triggered before language was loaded.

mediaDESIGN-SK:
Would it be possible to port this plugin to the CMS Joomla? I would take on this task.

florianbrinkmann:
Didn’t notice I was working in the wrong branch with the shell script modification 🙈 Should do no harm, so I leave it here.

PR #506, opened 2023-06-02
Antispam Bee 3

florianbrinkmann:


src/Handlers/Rules.php
```diff
@@ -0,0 +1,93 @@
...
+			if ( $rule_score > 0.0 ) {
+				$this->spam_reasons[] = $rule::get_slug();
+			} else {
+				$this->no_spam_reasons[] = $rule::get_slug();
+			}
+
+			$score += $rule_score;
+
+			$no_spam_threshold = (float) apply_filters( 'antispam_bee_no_spam_threshold', 0.0 );
+			$spam_threshold    = (float) apply_filters( 'antispam_bee_spam_threshold', 0.0 );
```
florianbrinkmann, line 36: Move this up before the foreach

src/Handlers/Rules.php
```diff
@@ -0,0 +1,93 @@
...
+
+			if ( $no_spam_threshold < 0.0 && $score <= $no_spam_threshold ) {
```
florianbrinkmann, line 38: Move the threshold checks below the `foreach`. Run the spam threshold check before the no_spam threshold check

src/load.php
```diff
@@ -0,0 +1,111 @@
...
+	foreach ( $modules as $module ) {
+		if ( is_callable( [ $module, 'always_init' ] ) ) {
+			call_user_func( [ $module, 'always_init' ] );
+		}
+
+		if ( $is_ajax_call && $disallow_ajax ) {
+			continue;
+		}
+
+		if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) {
```
2ndkauboy, line 96: Maybe also use a `$is_doing_autosave` variable above like for `$is_ajax_call`?

src/load.php
```diff
@@ -0,0 +1,111 @@
...
+use AntispamBee\Rules\ValidGravatar;
+
+/**
+ * Init function of the plugin
+ */
+function init() {
+	// Construct all modules to initialize.
+	$modules = [
+		DashboardWidgets::class,
+		new SettingsPage(),
```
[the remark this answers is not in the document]
florianbrinkmann: Because the `init` method isn’t static in that class

src/load.php
florianbrinkmann: Yep, good idea

src/Rules/Honeypot.php
```diff
@@ -0,0 +1,103 @@
...
+		
+		$request_uri  = Settings::get_key( $_SERVER, 'REQUEST_URI' );
+		$request_path = DataHelper::parse_url( $request_uri, 'path' );
+
+		if ( strpos( $request_path, 'wp-comments-post.php' ) === false ) {
+			return;
+		}
+		$fields = [];
+		foreach ( $_POST as $key => $value ) {
+			if ( isset( $fields['plugin_field'] ) ) {
```
florianbrinkmann, line 62: Here we rely on the order of the honeypot and the field that is replaced by it. We should not do that, and find a way to handle that independently from the order in `$_POST`

2ndkauboy: This get's fixed in https://github.com/pluginkollektiv/antispam-bee/pull/655

jmslbam:
Didn't now where to put the feedbac and maybe seperate issue, it's up to you, but here is my feedback https://toot.re/@2ndkauboy@dewp.space/116864017021004502
Danke!

2ndkauboy:
@jmslbam please feel free to open dedicated issues. Or just one with a collection of your findings. 

This PR will likely not be merged finally, as I think conserving the current version in the old `master` branch and using `v3` as the new development branch makes more sense.

PR #509, opened 2023-07-04
improve method to get client IP

florianbrinkmann:


pfefferle:
Any particular reason why you have removed the other checks?

It seems that you could get issues when using `REMOTE_ADDR` with a (reverse-)proxy. In these cases the `REMOTE_ADDR` could be the IP of the proxy.

stklcode:
I would not call this an “improvement“ for any use case.

Yes, there a a lot of potential issues arising from IP headers. But there are also many scenarios where we simply don’t get any better data, e.g. running a local apache/nginx reverse proxy stack (for whatever reason this is still quite common) or in cloud environments without direct access.

I would prefer

- making the trusted headers configurable, e.g. say we only trust the most common `X-Forwarded-For`
- reworking the logic, s.t. we take the first valid IP from the left of the list with fallback to `REMOTE_ADDR`

There are also implementations that add a list/subnet of trusted proxies, but sadly that can also be spoofed quite easily.

Simply dropping the headers may prevent us from parsing bogus, but also make the logic potentially useless in some deployments.

pfefferle:
Thanks for the clarification in slack.

Because it now uses the same functionality as the default WordPress implementation, I would remove the function/feature completely.

florianbrinkmann:
@stklcode making it configurable sounds good, do you have time to modify the PR?

pfefferle:
@stklcode do you have some time for a PR next week? Or can you at least explain your idea a bit more so maybe one of us can give it a try?

stklcode:
Like this commit: https://github.com/pluginkollektiv/antispam-bee/commit/282d6f9e1c743a80702a2e993c9d77f4a06daa3c in the [improve-ip-check-2](https://github.com/pluginkollektiv/antispam-bee/tree/improve-ip-check-2) branch.

I've done a quick implementation using a hook `antispam_bee_trusted_ip_headers` that receives an array of trusted headers with `['REMOTE_ADDR']` as default value and loops through it until a valid IP was found.
This might be migrated to a config flag in the UI with a comma-separated list or similar, if we feel the need for convenience here.

PR #513, opened 2023-07-20
only determine client IP from trusted headers

florianbrinkmann:
Code change and description by @stklcode

 By default, we only look for the client's IP address in the REMOTE_ADDR header. This can be extended via hook if there is a need to use headers like X-Forwarded-For ("HTTP_X_FORWARDED_FOR") instead.

Headers will be processed in the provided order and return the first valid IP address found.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -2423,7 +2423,7 @@ class Antispam_Bee {
 	 *
 	 * @since   2.6.1
 	 *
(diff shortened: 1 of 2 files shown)


Zodiac1978:
Maybe this filter is not necessary, as there is one in WP core since 1.5:
https://developer.wordpress.org/reference/hooks/pre_comment_user_ip/

Zodiac1978 (approved):
LGTM!

Issue #498, opened 2023-04-27
Add upgrade notification to plugin list screen


Zodiac1978:
The upgrade notice is only visible to the update page and not the plugin page.

Could be fixed with a filter. Here are two tutorials:
https://wisdomplugin.com/add-inline-plugin-update-message/
https://andidittrich.com/2015/05/howto-upgrade-notice-for-wordpress-plugins.html

Zodiac1978:
(Hopefully) fixed with #516

PR #516, opened 2023-07-21
show `upgrade_notice` also on plugin overview page

pfefferle:


pfefferle:
fixes #498

florianbrinkmann:
LGTM. PHPCS is complaining

pfefferle:
Yes, I think it's the autop! Should I remove it?

florianbrinkmann:
> Yes, I think it's the autop! Should I remove it?

Not sure, what do you mean @Zodiac1978 @2ndkauboy @krafit?

I think it would be great to be able to add links to the text, f.e. to a page with more details on pluginkollektiv.org, so maybe not strip all HTML but allow at least links.

pfefferle:
Ok, will change it to use kses instead

florianbrinkmann (approved):
LGTM

Issue #508, opened 2023-06-21
Warning in the error log


Zodiac1978:
**Describe the bug**

There are php warning the error log with PHP 8.1 (first reported in the [forums](https://wordpress.org/support/topic/php-8-1-notices-3/))

```
    [20-Jun-2023 20:22:23 UTC] PHP Warning: Trying to access array offset on value of type null in /home/wp-content/plugins/antispam-bee/antispam_bee.php on line 2626
    [20-Jun-2023 20:22:23 UTC] PHP Warning: Undefined array key “reasons” in /home/wp-content/plugins/antispam-bee/antispam_bee.php on line 2626
```

Here is a link to the line:
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L2626

I'm afraid this is another consequence of moving the defaults to its own function in https://github.com/pluginkollektiv/antispam-bee/pull/478/commits/2cd7a991c973dc26474841c5fccd66a2b9ddd19b

@florianbrinkmann Can you have a look at this?

florianbrinkmann:
@Zodiac1978 I created a PR that should fix the issue (#517)

Zodiac1978:
Fixed with #517

PR #517, opened 2023-07-26
fix: populate reasons array before sending spam notice mail, refs #508

florianbrinkmann:


pfefferle:
SonarCloud is throwing an error, we should either fix it or remove SonarCloud completely in favor of the local tests.

florianbrinkmann:
As far as I see, the SonarCloud issue is about the SQL query in the `_is_db_spam` method, no idea why it pops up now (or I miss something in the SonarCloud report)

stklcode:
This one is mostly a false positive imho.

Sonarqube points out a 15 step chain very well: a user can craft an HTTP request with malicious IP headers that are put into a generated SQL statement.

But this specific “user input“ will run through `_sanitize_ip()` which will not return any non-ip characters and thus won’t allow an SQL injection through IP headers.

If `filter_var` does not exist (is that the case with any PHP setup?), our fallback would allow values of invalid length and might open an attack vector for overflow/performance issues. But I guess this is more of a highly theoretical scenario.

Zodiac1978 (approved):
LGTM!

PR #529, opened 2023-09-20
Fix empty data error

### LINKED ISSUE #527: Usage of pre_comment_user_ip is breaking if IP is removed
If the IP is removed (via snippet or GDPR plugin) our "Empty data" check is triggered:

https://github.com/pluginkollektiv/antispam-bee/blob/ab767d77b6406e25f7209ae8a861452c305f712e/antispam_bee.php#L1462-L1466

First reported here: https://wordpress.org/support/topic/comments-are-marked-as-spam-reason-empty-data/

So we need to use our own filter after all and can't use the core filter. Or we do need to remove this check.

Zodiac1978:
Usage of the core filter pre_comment_user_ip does trigger our empty data check for the IP address if the IP is removed intentionally for GDPR compliance. Switching to own filter.

Fixes #527

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -2458,7 +2458,7 @@ class Antispam_Bee {
 	 */
 	public static function get_client_ip() {
 		/**
-		 * WordPress hook for allowing to modify the client IP used by Antispam Bee. Default value is the `REMOTE_ADDR`.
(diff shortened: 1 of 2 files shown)


florianbrinkmann (approved):
LGTM

PR #550, opened 2024-01-07
Delete missed out option on uninstall

### LINKED ISSUE #549: Option is left in DB after uninstalling (and removing data checked)
Reported in the forums:
https://wordpress.org/support/topic/did-not-delete-some-data-in-db-on-uninstall/

The option `antispambee_db_version` is not part of our `antispam_bee` options array, but an own entry.

In our uninstall function, we missed deleting this option.

Zodiac1978:
Fixes #549


stklcode (approved):
LGTM

PR #564, opened 2024-02-03
Better support for AJAX comments

### LINKED ISSUE #149: Allow creation of comment form via ajax
Hi everyone!

I just want to understand your decision to block all AJAX calls in the plugin init() function
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L76

In our case we just want to create the comment form via AJAX because we are working with a page cache and the comment form will be loaded in a popup after clicking the "Add a comment" Button

Before I work out a pull request to work around this issue I just want to understand the decision to block all AJAX calls.

Thanks :)

Zodiac1978:
Ongoing fixes for better AJAX comments support

See: https://wpdirectory.net/search/01HJKQW4Q7NCTKNB3W54CPARMZ

TODO: https://github.com/pluginkollektiv/antispam-bee/issues/149#issuecomment-1870265781

Fixes #149

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -165,7 +165,7 @@ class Antispam_Bee {
 				)
 			);
 
-		} elseif ( is_admin() ) {
+		} elseif ( is_admin() && ! defined( 'DOING_AJAX' ) ) {
 			add_action(
 				'admin_menu',
 				array(


stklcode:
This looks good to me, so I tend to merge it as is.

I'm not too deep into the inner logic right now, but I do not see this doing harm either, so one step into the right direction.

Zodiac1978:
This is the first step to fix #149 

There is another fix needed.
https://github.com/pluginkollektiv/antispam-bee/issues/149#issuecomment-1870265781

But it would help to avoid unnecessary code execution for AJAX calls - but as this wouldn't make much sense without fixing the above this PR could wait until we fix the other problem.

PR #565, opened 2024-02-04
Better comment type handling

MatzeKitt:
- Process WooCommerce reviews by default
- Allow filtering processable comment types

### DIFF
--- a/src/Handlers/Comment.php
+++ b/src/Handlers/Comment.php
@@ -23,7 +23,7 @@ class Comment extends Reaction {
 	}
 
 	public static function process( $comment ) {
-		if ( ! ContentTypeHelper::reaction_is_one_of( $comment, [ 'comment', '' ], 'comment' ) ) {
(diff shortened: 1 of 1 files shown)


src/Handlers/Comment.php
```diff
@@ -23,7 +23,14 @@ function () {
 	}
 
 	public static function process( $comment ) {
-		if ( ! ContentTypeHelper::reaction_is_one_of( $comment, [ 'comment', '' ], 'comment' ) ) {
+		/**
+		 * Filter processable comment types.
+		 *
+		 * @param	array	$types List of comment types
+		 */
+		$comment_types = apply_filters( 'antispam_bee_comment_types', [ '', 'comment', 'review' ] );
```
florianbrinkmann, line 31: Maybe add a `(array)` here to ensure we get an array back?

florianbrinkmann (approved):
LGTM!

PR #573, opened 2024-03-10
Create blueprint.json

### LINKED ISSUE #557: Add blueprint.json to support playground
We could add a blueprint.json to support the playground.

For example, to run some custom PHP to insert (spam) comments to better show our features:
https://wordpress.github.io/wordpress-playground/blueprints-api/examples#run-custom-php-code

See also:
https://joost.blog/plugin-demos-with-the-wordpress-playground/

Or to add some pointers in the playground:
https://www.alexgeorgiou.gr/wordpress-admin-pointers-for-dummies/

Zodiac1978:
Add a basic blueprint.json to show our settings page.

Additionally, I also set up a constant to detect the playground for further tweaking.

As an alternative we could show the spam folder, but this would need some spam examples we need to import or add to the database, which needs more work. As a first step, this could also be helpful.

#557

### DIFF
--- a/.wordpress-org/blueprint/blueprint.json
+++ b/.wordpress-org/blueprint/blueprint.json
@@ -1,0 +1,16 @@
+{
+	"landingPage": "\/wp-admin\/options-general.php?page=antispam_bee",
+	"steps": [
+		{
+			"step": "login",
+			"username": "admin",
+			"password": "password"
+		},
+		{
+			"step": "defineWpConfigConsts",
+			"consts": {
+			  "IS_PLAYGROUND_PREVIEW": true
+			}
+		}
+	]
+}


florianbrinkmann:
I think showing the settings is a good first step

.wordpress-org/blueprint/blueprint.json
```diff
@@ -0,0 +1,16 @@
...
+	"steps": [
+		{
+			"step": "login",
+			"username": "admin",
+			"password": "password"
+		},
+		{
+			"step": "defineWpConfigConsts",
+			"consts": {
+			  "IS_PLAYGROUND_PREVIEW": true
```
florianbrinkmann, line 12: Not sure if we should prefix that, to make it less likely that the constant is added by something else

florianbrinkmann (approved):

.wordpress-org/blueprint/blueprint.json
Zodiac1978: This is only added on Playground. Network is not activated, so you can't add another plugin.

PR #576, opened 2024-03-17
[v3] fix: add missing default return value to EmptyData rule

stklcode:


florianbrinkmann (approved):
Oh, good catch, thanks!

PR #578, opened 2024-03-17
[v3] restore unit test environment

stklcode:
Update test dependencies to restore PHPUnit tests.
I didn't care about the old Behat stuff and commented out the pipeline part.

Started with one random test for `EmptyData` rule and already found (and fixed) a real bug :smiley:


florianbrinkmann:
> Started with one random test for EmptyData rule and already found (and fixed) a real bug 😃

Whoop! 🎉

florianbrinkmann (approved):

PR #579, opened 2024-03-17
[v3] minor fixes, code style tweaks, extend phpdocs

stklcode:
Working through all PHPCS issues. Discovered some minor issues, but most issues are just comments and documentation.

Yes, some comments are pure repetition without much of additional information, but it satisfies the code-style checks for now. We might spent some time in the future to improve on that, if necessary.


florianbrinkmann (approved):
Thanks a lot! ❤️

PR #589, opened 2024-05-06
Use `SCRIPT_NAME` instead of `REQUEST_URI` to check path

### LINKED ISSUE #585: Hide My WP Ghost - Security Obfuscation
**Describe the bug**
Getting the dreaded" No more “Error: Please type your comment text.” "
This is due to the wp-comments-post.php is hard coded in the php code 
The security plugin I use changes to whatever I want such as "x89yh8g1273g48yhj".
It's useful because it makes it harder for attackers to find the right string to attack.

**To Reproduce**
1. Install Hide My WP Ghost 
2. Obscure all the default paths including wp-comments-post.php
3. Install Antispam Bee
4. Can't post any comments, get the following: Error: Please type your comment text.”

**Expected behavior**
I expect to be able to post.

- OS: ALL
- Version Antispam Bee version 2.11.6

**My original fix for myself**
Posted this 1:1 on wordpress.org, hope you can make an option to set custom wp-comments-post.php

I wanted to contribute a fix to make this work with Hide My WP Plugin.
Open the zip file using file explorer (NOT WINRAR)
Copy to local antispam_bee.php
modify the wp-comments-post.php
Change it to what you currently have “comments “Custom comment Path:” set
I have something like “x89yh8g1273g48yhj”
There are 2 places you need to modify line 1185 and 1140.
This is for the version 2.11.6 of Antispam Bee.
Delete the original antispam-bee.2.11.6\antispam-bee\antispam_bee.php
Put your modified antispam_bee.php in that directory.
No more “Error: Please type your comment text.”
I hope this helps someone out there:)

2ndkauboy:
The script is currently checking if the `REQUEST_URI` is containing `wp-comments-post.php`, the default script to handle the submission of a comment. Some security plugins have options to rename this file to disguise that WordPress is used.

With this fix, the `SCRIPT_NAME` is used instead. Since many security plugins do use rewrite rules, while the `REQUEST_URI` value is changed, the `SCRIPT_NAME` value stays the same. Therefore, the condition would still recognize if a comment was submitted.

Fixes #585

### DIFF
--- a/tests/Unit/AntispamBeeTest.php
+++ b/tests/Unit/AntispamBeeTest.php
@@ -68,7 +68,7 @@ class FactoryTest extends TestCase {
 		$_SERVER['REMOTE_ADDR']          = '192.0.2.1';
 		$_SERVER['HTTP_X_FORWARDED_FOR'] = '192.0.2.2, 10.0.0.10';
 		$_SERVER['HTTP_X_REAL_IP']       = 'bogus';
-		$_SERVER['REQUEST_URI']          = 'https://domain.com/wp-comments-post.php';
+		$_SERVER['SCRIPT_NAME']          = '/wp-comments-post.php';
 		$_POST['comment']                = $comment;
 
 		$result = Testee::handle_incoming_request( $comment );
@@ -99,7 +99,7 @@ class FactoryTest extends TestCase {
 		$comment = array_merge( $this->get_base_comment(), $comment );
 
 		$_SERVER['REMOTE_ADDR'] = '12.23.34.45';
-		$_SERVER['REQUEST_URI'] = 'https://domain.com/wp-comments-post.php';
+		$_SERVER['SCRIPT_NAME'] = '/wp-comments-post.php';
 		$_POST['comment']       = $comment;
 
 		// This is where we check for the spam reason that was detected.
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1134,7 +1134,7 @@ class Antispam_Bee {
 			return;
 		}
 
-		$request_uri  = self::get_key( $_SERVER, 'REQUEST_URI' );
+		$request_uri  = self::get_key( $_SERVER, 'SCRIPT_NAME' );
 		$request_path = self::parse_url( $request_uri, 'path' );
 
 		if ( strpos( $request_path, 'wp-comments-post.php' ) === false ) {
@@ -1168,7 +1168,7 @@ class Antispam_Bee {
 	public static function handle_incoming_request( $comment ) {
 		$comment['comment_author_IP'] = self::get_client_ip();
 
-		$request_uri  = self::get_key( $_SERVER, 'REQUEST_URI' );
+		$request_uri  = self::get_key( $_SERVER, 'SCRIPT_NAME' );
 		$request_path = self::parse_url( $request_uri, 'path' );
 
 		if ( empty( $request_path ) ) {


tests/Unit/AntispamBeeTest.php
```diff
@@ -68,7 +68,7 @@ public function test_gets_ip_address() {
 		$_SERVER['REMOTE_ADDR']          = '192.0.2.1';
 		$_SERVER['HTTP_X_FORWARDED_FOR'] = '192.0.2.2, 10.0.0.10';
 		$_SERVER['HTTP_X_REAL_IP']       = 'bogus';
-		$_SERVER['REQUEST_URI']          = 'https://domain.com/wp-comments-post.php';
+		$_SERVER['SERVER_NAME']          = 'https://domain.com/wp-comments-post.php';
```
stklcode, line 71: Guess you wanted to use `SCRIPT_NAME` here?

This car typically does not contain the full URL, so a more realistic value would be
```php
$_SERVER['SCRIPT_NAME']          = '/wp-comments-post.php';
```

2ndkauboy: Right!

2ndkauboy:
Somehow, I accidentally closed the PR 🤔

tests/Unit/AntispamBeeTest.php
2ndkauboy: But the `REQUEST_URI` would often be the same, so the code was wrong before 😅 

@stklcode do you know why the tests are broken?

2ndkauboy: I think I've found the issue 🙈 The error message wasn't really of big help 😞

2ndkauboy: @stklcode any idea on why the Behat tests are failing?

stklcode: Unfortunately no. I'm not really familiar with _Behat_ and the bootstrap script is apparently only just running (behat.sh even throws some errors that are ignored) in GH actions, so it's little tricky to debug.

But the failure seems unrelated to your changes. I reverted everything temporarily and CI still fails: https://github.com/pluginkollektiv/antispam-bee/actions/runs/9066345933/job/24909046007?pr=589

stklcode: Likely a currently broken language API.

2ndkauboy: So you would say we can merge this PR?

stklcode: Yes. Changes are reasonable and look good.

... whenever GH actions finally reports the status, system is degraded right now.

stklcode (approved):

2ndkauboy:
@Zodiac1978 I don't think it's an alternative approach, it's just another step. Currently, the condition is the following:

```
strpos( $request_path, 'wp-comments-post.php' ) !== false
```

The `$request_path` used to be the `REQUEST_URI`, which would change, if a security plugin "changes this script name" using a rewrite rule. By using `SCRIPT_NAME`, we make sure we get the script name of the actual script that handles the request.

This is only affecting the left side of the condition and the first parameter `$request_path`. With `home_url()`, we could replace the second parameter, the static `wp-comments-post.php` string. But this would still not solve the issue. Replacing the first parameter with `home_url()`, the condition would most probably always be true, right?

Since this PR resolves the issue, I would not introduce a new function call (twice), we probably don't need.

stklcode:
`home_url(…)` should equal `REQUEST_URI`. But I don’t see much of a benefit over the (cheaper) `SCRIPT_NAME` comparison.

2ndkauboy:
The `home_url()` path is not looking at any `$_SERVER` values at all. All it does (in this case) is prepending the path with a slash. So calling `home_url('wp-comments-post.php', 'relative')` does always return `/wp-comments-post.php`, unless there is a filter hooking into `home_url`, which Hide My WP does not seem to do. So this is no solution.

@Zodiac1978 are you also OK to merge this then?

Zodiac1978:
> are you also OK to merge this then?

I just asked the author and if his approach is not smart, let's do it better.

From my understanding, it would still solve the issue with his plugin, so yes, I'm fine with PR!

Zodiac1978 (approved):
LGTM!

PR #593, opened 2024-05-16
fix: use SCRIPT_NAME instead of REQUEST_URI to check path (#585)

stklcode:
The script is currently checking if the `REQUEST_URI` is containing `wp-comments-post.php`, the default script to handle the submission of a comment. Some security plugins have options to rename this file to disguise that WordPress is used.

With this fix, the `SCRIPT_NAME` is used instead. Since many security plugins do use rewrite rules, while the `REQUEST_URI` value is changed, the `SCRIPT_NAME` value stays the same. Therefor the condition would still recognize if a comment was submitted.

Original fix by @2ndkauboy in #589, adapted to v3.

### DIFF
--- a/src/Rules/Honeypot.php
+++ b/src/Rules/Honeypot.php
@@ -81,7 +81,7 @@ class Honeypot extends ControllableBase implements SpamReason {
 			return;
 		}
 
-		$request_uri  = Settings::get_key( $_SERVER, 'REQUEST_URI' );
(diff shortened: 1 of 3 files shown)


2ndkauboy (approved):

2ndkauboy:
@stklcode sorry that I've missed this already existing PR. Yours fixes one more place I've missed. And it adds test! :raised_hands: 

So please feel free to merge it now.

PR #595, opened 2024-05-22
Allow change of debug log directory

florianbrinkmann:


src/Helpers/DebugMode.php
```diff
@@ -42,11 +42,16 @@ public static function log( string $message ) {
 			return;
 		}
 
-		$date        = date( 'Y-m-d' ); // phpcs:ignore WordPress.DateTime.RestrictedFunctions.date_date
-		$time        = date( 'H-i-s' ); // phpcs:ignore WordPress.DateTime.RestrictedFunctions.date_date
-		$content_dir = \WP_CONTENT_DIR;
+		$date    = date( 'Y-m-d' ); // phpcs:ignore WordPress.DateTime.RestrictedFunctions.date_date
+		$time    = date( 'H-i-s' ); // phpcs:ignore WordPress.DateTime.RestrictedFunctions.date_date
+		$log_dir = defined( 'ANTISPAM_BEE_DEBUG_MODE_LOG_DIR' ) ? \ANTISPAM_BEE_DEBUG_MODE_LOG_DIR : \WP_CONTENT_DIR;
```
stklcode, line 47: Pretty long const name, would have preferred something a little shorter, e.g. strip the `_MODE`. But we should probably keep it consistent with `ANTISPAM_BEE_DEBUG_MODE_ENABLED`, so I'm fine with either.

stklcode (approved):

src/Helpers/DebugMode.php
florianbrinkmann: True, it’s a little long 😬 Because we are not live yet and only two or three people might be using it, I would say let’s change the `ANTISPAM_BEE_DEBUG_MODE_ENABLED` to `ANTISPAM_BEE_DEBUG` and the log to `ANTISPAM_BEE_DEBUG_DIR`?

2ndkauboy: How about `ANTISPAM_BEE_DEBUG` and `ANTISPAM_BEE_DEBUG_LOG`, following the naming of the WordPress constants like `WP_DEBUG_LOG`?

2ndkauboy:
Cross-referencing: the constant naming discussed in the review here has grown a second half. #797 adds reason slugs to the Fail2Ban spam log, and one option for keeping existing jails working is to introduce a **new** constant for the new log rather than a format switch on the old one — which means we would be naming two constants, not one, and they should agree with each other.

I have written up the details in [this comment on #797](https://github.com/pluginkollektiv/antispam-bee/pull/797#issuecomment-5228476063). Short version, extending @2ndkauboy's `WP_DEBUG` / `WP_DEBUG_LOG` suggestion to cover both:

* `ANTISPAM_BEE_DEBUG` — enable debug logging
* `ANTISPAM_BEE_DEBUG_LOG` — where the debug log goes
* `ANTISPAM_BEE_SPAM_LOG` — where the spam log goes, replacing `ANTISPAM_BEE_LOG_FILE`

Two things specific to this PR:

**`WP_DEBUG_LOG` takes a file path, not a directory.** So naming ours `ANTISPAM_BEE_DEBUG_LOG` for familiarity invites people to pass a file — which would defeat the salted filename `DebugMode` generates. That salt matters: the debug log contains comment data and `WP_CONTENT_DIR` is usually served publicly, so given the choice people will pick `debug.log` and publish it. The directory that @florianbrinkmann's diff already uses avoids this, and `ANTISPAM_BEE_DEBUG_LOG_DIR` states it plainly. The alternative is to accept a file path and append the salted suffix anyway, but then the name promises more control than it gives.

**This branch predates the salt.** The diff writes `asb-debug.{$date}.log`, while `v3` now writes `asb-debug.{$date}.{$suffix}.log` via `get_log_file_suffix()`. It needs a rebase, and the suffix should survive it — a straight replay of the current diff would silently drop the protection.

No objection to the change itself; it is only the naming that is worth settling in one place, since #797 will otherwise pick a name independently.

PR #596, opened 2024-05-22
Add `tests` directory to phpcs check

florianbrinkmann:


2ndkauboy:
We should totally do that. And also resolve all warnings in the current code. But maybe not for the `3.0.0-beta.1` :see_no_evil:

PR #607, opened 2024-07-06
Add `scrape` string to useragent regex

florianbrinkmann:


2ndkauboy:
@florianbrinkmann can you explain what type of scraper/bot this detects? I could also not find it in `v2`. So maybe something for a future release?

PR #624, opened 2024-10-14
Add filter to change honeypot styles

2ndkauboy:
This filter can be used to change the styles of the honeypot field. It can also use an empty value, and then load the styles from an external file, if a strict CSP is used for the site.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1305,11 +1305,25 @@ class Antispam_Bee {
 			}
 		}
 
+		/**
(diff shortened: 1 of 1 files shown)


2ndkauboy:
@Zodiac1978 can you approve this PR? I can't run the tests myself, since the PR needs to be approved, which I can't do, since I've opened the PR.

Zodiac1978 (approved):
LGTM!

PR #627, opened 2024-10-18
Remove `add_reasons_to_defaults()` from general initialization

### LINKED ISSUE #625: Notice: Function _load_textdomain_just_in_time was called incorrectly
**Describe the bug**

_Notice: Function _load_textdomain_just_in_time was called incorrectly. Translation loading for the antispam-bee domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the init action or later. Please see Debugging in WordPress for more information. (This message was added in version 6.7.0.) in /home/mkjqekzy/public_html/wp-includes/functions.php on line 6099_

**To Reproduce**

1. Install WordPress beta tester plugin
2. Update to WordPress 6.7 beta 3
3. Enable WP_DEBUG
4. Enable AntiSpam Bee

**Expected behavior**
It looks like a deprecation warning of some sort. Not that unexpected really ;)

- WordPress 6.7 beta 3
- Firefox browser
- PHP 8.2.x

**Additional context**
image

2ndkauboy:
The `add_reasons_to_defaults()` adds the translatable strings to the `$default`. Those `reasons` are only needed inside the `send_mail_notification()` method, which also calls the `add_reasons_to_defaults()` function. Therefore, the function call can be removed from the general `init()` function.

Fixes #625

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -151,11 +151,6 @@ class Antispam_Bee {
 
 		self::_init_internal_vars();
 
-		if ( self::_current_page( 'dashboard' ) || self::_current_page( 'plugins' ) || self::_current_page( 'options' ) || self::_current_page( 'edit-comments' ) || self::_current_page( 'admin-post' ) ) {
-			self::load_plugin_lang();
-			self::add_reasons_to_defaults();
-		}
-
 		if ( defined( 'DOING_CRON' ) ) {
 			add_action(
 				'antispam_bee_daily_cronjob',
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -151,6 +151,13 @@ class Antispam_Bee {
 
 		self::_init_internal_vars();
 
+		add_action(
+			'init', array(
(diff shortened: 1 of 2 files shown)


swissspidy:
I think this re-introduces #508.

1. Go to the Antispam Bee settings page
2. See warnings

Query Monitor screenshot:

Screenshot 2024-10-18 at 14 43 53

----

Translation-wise it looks fine though.

2ndkauboy:
@swissspidy great catch!

I first wondered on how/where the `Antispam_Bee_GUI` was able to get `self::$defaults`, but then I saw it: `... extends Antispam_Bee` :see_no_evil: 

Oh boy! Time to get this old code gone and `v3` ready!

My new commit should fix it. It now loads the "reasons" in an `init` hook callback. Still too much, but it should resolve both issues.

swissspidy (approved):

stklcode (approved):
LGTM

PR #655, opened 2025-07-27
Get hidden and plugin field correctly

2ndkauboy:
The new code is using a `foreach` to get the two fields, but requires them to be in the correct order.

This change uses a code similar to v2 which would get the parameters, independent of the order.

Fixes: https://github.com/pluginkollektiv/antispam-bee/pull/506#pullrequestreview-1837201864

### DIFF
--- a/src/Rules/Honeypot.php
+++ b/src/Rules/Honeypot.php
@@ -88,32 +88,18 @@ class Honeypot extends ControllableBase implements SpamReason {
 		if ( strpos( $request_path, 'wp-comments-post.php' ) === false ) {
 			return;
 		}
-		$fields = [];
-		foreach ( $_POST as $key => $value ) {
-			if ( isset( $fields['plugin_field'] ) ) {
-				$fields['hidden_field'] = $key;
-				break;
-			}
-			if ( HoneypotField::get_secret_name_for_post() === $key ) {
-				$fields['plugin_field'] = $key;
-			}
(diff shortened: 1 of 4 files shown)


florianbrinkmann (approved):

src/Rules/Honeypot.php
```diff
@@ -87,32 +87,18 @@ public static function precheck(): void {
...
-				break;
-			}
-			if ( HoneypotField::get_secret_name_for_post() === $key ) {
-				$fields['plugin_field'] = $key;
-			}
-		}
 
-		if ( ! isset( $fields['plugin_field'] ) ) {
-			// Honeypot field was not present in $_POST data.
-			$_POST['ab_spam__invalid_request'] = 1;
```
stklcode, line 103: There is a test case for this field that now fails:
https://github.com/pluginkollektiv/antispam-bee/blob/78e84de0d1eee3a9b2a670cf04921fdd5ba4fad7/tests/Unit/Rules/HoneypotTest.php#L72-L74

Should be either adjusted or removed if obsolete.

2ndkauboy: This test was removed. I've also changed the field names in the tests, since they are currently not dynamic in the way proposed in the tests. The hidden field is always `comment` for this comment form. This is currently hardcoded in our plugin (as it was in v2).

PR #656, opened 2025-08-07
Add `ab_spam__invalid_request` logic back to `precheck()` function

2ndkauboy:
@stklcode thanks for your comment on #655 about a failing test. After fixing the implementation of the Honypot precheck, I've restored the logic from `v2`. In my opinion, the initial implementation of this rule in `v3` added an additional validation step, which might change how spam is detected. I believe it's a good addition, but I'm not sure it we should include it without further testing.

This PR restored that new logic and also adds back the test that was failing in #655, which now works again.


2ndkauboy:
New tests are added. This is also a great thing to have beta testers for. :sweat_smile:

PR #658, opened 2025-09-07
Looser regex for non-quoted attributes

### LINKED ISSUE #657: Regex doesn't pick up forms without quoted attributes.
**Describe the bug**
As discussed in https://wordpress.org/support/topic/honeypot-and-use-with-comment_form/

My comment box's HTML looks like:

`<textarea name=comment …>`

This means AntiSpam Bee doesn't detect it and add the honeypot.

**Expected behavior**
The regex should pick up unquoted attributes.

**Additional context**

I got it working by changing the regex to:

```php
if ( ! preg_match( '#<textarea.+?name=["\']?comment["\']?#s', $data ) ) {
        return $data;
}

return preg_replace_callback(
        '/(?P<all>                                                              (?# match the whole textarea tag )
                <textarea                                                           (?# the opening of the textarea and some optional attributes )
                (                                                                   (?# match a id attribute followed by some optional ones and the name attribute )
                        (?P<before1>[^>]*)
                        (?P<id1>id=["\'](?P<id_value1>[^>"\']*)["\'])
                        (?P<between1>[^>]*)
                        name=["\']?comment["\']?
                        |                                                               (?# match same as before, but with the name attribute before the id attribute )
                        (?P<before2>[^>]*)
                        name=["\']?comment["\']?
                        (?P<between2>[^>]*)
                        (?P<id2>id=["\'](?P<id_value2>[^>"\']*)["\'])
                        |                                                               (?# match same as before, but with no id attribute )
                        (?P<before3>[^>]*)
                        name=["\']?comment["\']?
                        (?P<between3>[^>]*)
                )
                (?P<after>[^>]*)                                                    (?# match any additional optional attributes )
                >                                                                   (?# the closing of the textarea opening tag )
                (?s)(?P<content>.*?)                                                (?# any textarea content )
                <\/textarea>                                                        (?# the closing textarea tag )
        )/x',
        array( 'Antispam_Bee', 'replace_comment_field_callback' ),
        $data,
        -1
);
```

Would you like me to send a PR for that?

Longer term, it might be worth using https://www.php.net/manual/en/class.dom-htmldocument.php to find and replace the elements.

edent:
Fixes #657

(Sorry for the early snafu!)

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1232,7 +1232,7 @@ class Antispam_Bee {
 			return $data;
 		}
 
-		if ( ! preg_match( '#<textarea.+?name=["\']comment["\']#s', $data ) ) {
+		if ( ! preg_match( '#<textarea.+?name=["\']?comment["\']?#s', $data ) ) {
 			return $data;
 		}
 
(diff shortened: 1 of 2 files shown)


stklcode (approved):
LGTM, thanks for the patch.

Just added a short test case.

PR #676, opened 2026-03-14
[v3] Use looser regex for non-quoted attributes (#657)

stklcode:
Markup preparation for textarea elements with unquoted "name" attribute are not properly detected and are not augmented by Antispam Bee.

Loosen the expression to also match unquoted attributes which are valid HTML and should be processed equally.

Adapted to v3 from: 4e36755df73cba0942df28d761d7ee447f12b9d3 (#658)


stklcode:
superseded by #731

PR #686, opened 2026-04-19
Add shorturl.fm spam pattern

### LINKED ISSUE #685: New Regex Pattern
**Is your feature request related to a problem? Please describe.**
I recently got hit with a wave of shorturl spam comments on my website. They all followed a simple pattern. A comment that contains nothing but a URL starting with either http:// or https://, followed by shorturl.fm/, followed by exactly 5 alphanumeric characters.

**Describe the solution you'd like**
I would like to add a regex to prevent these comments.
/^https?:\/\/shorturl\.fm\/[a-zA-Z0-9]{5}$/

**Describe alternatives you've considered**
I considered using **/S** instead of **a-zA-Z0-9** however that includes punctuation which is not necessary in the context.

The comments only used **https** but I included **http** as well. 

I considered having this prevent any comment that includes a shorturl link however that would be unreasonable. If two real users were engaged in conversation it is not unlikely that one of them shares a link.

**Additional context**
Here is one example of a spam comment that matches the pattern I described above. This is a weak pattern and can be prevented.
Image

LIoccoUMD:
Closes #685 

Adds a body pattern to block spam comments consisting solely of a shorturl.fm URL with a 5-character alphanumeric slug.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1655,6 +1655,9 @@ class Antispam_Bee {
 				'body'  => 'dating|sex|lotto|pharmacy',
 				'email' => '@mail\.ru|@yandex\.',
 			),
+			array(
+				'body' => '^https?:\/\/shorturl\.fm\/[a-zA-Z0-9]{5}$',
+			),
 		);
 
 		$quoted_author = preg_quote( $comment['author'], '/' );
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1657,6 +1657,8 @@ class Antispam_Bee {
 			),
 			array(
 				'body' => '^https?:\/\/shorturl\.fm\/[a-zA-Z0-9]{5}$',
+				'email' => '@gmail\.com',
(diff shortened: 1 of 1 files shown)


antispam_bee.php
```diff
@@ -1655,6 +1655,9 @@ private static function _is_regexp_spam( $comment ) {
 				'body'  => 'dating|sex|lotto|pharmacy',
 				'email' => '@mail\.ru|@yandex\.',
 			),
+			array(
+				'body' => '^https?:\/\/shorturl\.fm\/[a-zA-Z0-9]{5}$',
```
2ndkauboy, line 1659:
```suggestion
				'body' => '^https?:\/\/shorturl\.fm\/[a-zA-Z0-9]{5}$',
				'email' => '@gmail\.com',
```

Maybe also add the `author` pattern.

2ndkauboy (approved):

PR #693, opened 2026-05-29
fix: return bool from _is_mobile() instead of raw strpos result

### LINKED ISSUE #670: _is_mobile() returns raw strpos result, not a strict boolean
If match occurs at position x (int), the function returns x (and not true/false), which is a logic bug:
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L2258-L2268

It works, because PHP interprets 0 as false and any int as true.

thisismyurl:
`strpos()` returns `int|false`, so `_is_mobile()` was returning an integer position when wptouch was found rather than the `bool` its PHPDoc promises. This just adds `false !==` to make the return type correct.

Fixes #670.

Built and tested with AI assistance.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -2271,7 +2271,7 @@ class Antispam_Bee {
 	 * @return  boolean  TRUE if "wptouch" is active
 	 */
 	private static function _is_mobile() {
-		return strpos( get_template_directory(), 'wptouch' );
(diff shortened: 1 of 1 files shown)


2ndkauboy:
Thanks for this PR! Looks good.

2ndkauboy (approved):

PR #692, opened 2026-05-28
fix: pass integer to _n() in dashboard spam counter

robertstaddon:
## Summary

Fixes a PHP fatal `TypeError` on the WordPress Dashboard when the dashboard spam counter option is enabled on sites with a strictly typed `ngettext` filter (reported with Divi 5.6 / `D4DomainBridge`).

`add_dashboard_count()` passed `self::_get_spam_count()` as the third argument to `_n()`. That method returns a locale-formatted string (`number_format()` / `number_format_i18n()`), but `_n()` expects a numeric value for singular/plural selection only. The formatted value should be used only for display in `sprintf()`.

## Changes

- Pass `intval( self::get_option( 'spam_count' ) )` to `_n()`
- Continue using `self::_get_spam_count()` for the formatted count in `sprintf()`

## How to test

1. Enable **Display spam counter on dashboard** in Antispam Bee settings.
2. Ensure the spam count is greater than 0.
3. Open **Dashboard** as an administrator.
4. Confirm the blocked count displays and no PHP fatal appears in `debug.log`.

Recommended: test on PHP 8+ with Divi 5.6 if available (environment that previously reproduced the error).

## Related

- https://wordpress.org/support/topic/dashboard-fatal-_n-given-formatted-string-instead-of-integer/
- Fixes https://github.com/pluginkollektiv/antispam-bee/issues/691

PR #694, opened 2026-05-29
refactor: improve number formatting and pluralization in dashboard widgets

2ndkauboy:
Replace custom number formatting with WordPress i18n functions and add proper singular/plural forms for spam statistics. Remove German-specific number formatting in favor of locale-aware number_format_i18n(). Simplify get_spam_count() to return an integer and move formatting to the display layer.

This ports #692 to https://github.com/pluginkollektiv/antispam-bee/tree/v3

### DIFF
--- a/src/Admin/DashboardWidgets.php
+++ b/src/Admin/DashboardWidgets.php
@@ -41,23 +41,41 @@ class DashboardWidgets {
 
 		echo '<style>#dashboard_right_now .ab-count::before {content: "\f117"} #dashboard_right_now .ab-current-spam::before {content: "\f17e"}</style>';
 
-		$items[] = '<span class="ab-count">' . esc_html(
-			sprintf(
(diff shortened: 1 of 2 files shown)


2ndkauboy:
I'm wondering if we should also delete the `the_spam_count()` function altogether. It is only used in the `antispam_bee_count` which the codebase does not use. It was documented in the past so that users could do something like `do_action('antispam_bee_count')` to print the spam count somewhere (which was necessary since all functions are private in `v2`).

Should we have some kind of `deprecated.php` file where we move actions/filters/functions/options that we don't need in the new version but to not break current code using them?

2ndkauboy:
Since the `the_spam_count()` function was never part of a "public API" and it only seems to be [used in three plugins](https://veloria.dev/search/e0946231-3af9-4dae-ad92-b65e33b7d673), all having a check around, we can remove it. It wouldn't work anyway, since it does not exist in the same class.

PR #698, opened 2026-06-07
Fix upgrade_notice(): wrong hook tag (absolute path vs. plugin basename) and trailing space in array key

### LINKED ISSUE #669: Upgrade notice likely never rendered correctly
Hook uses absolute path instead of plugin basename
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L3014

Notice body reads wrong key with trailing space
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L1096

(Found via Codex)

Ping @pfefferle as you implemented this feature

thisismyurl:
Fixes #669.

Two bugs in `upgrade_notice()` that together mean this feature has never worked for any Antispam Bee release.

**Bug 1 — Wrong hook tag:** `add_action( 'in_plugin_update_message-' . __FILE__, ... )` — `__FILE__` resolves to the absolute server path. WordPress registers the hook under the plugin basename (`antispam-bee/antispam_bee.php`), so the action was never matched and `upgrade_notice()` was never called. Fix: `plugin_basename( __FILE__ )`.

**Bug 2 — Trailing space in array key:** `$data['upgrade_notice ']` — the key has a trailing space. Even if the hook had fired, `isset()` would always fail because WordPress passes `'upgrade_notice'` (no space). Fix: remove the trailing space.

To verify: trigger the in_plugin_update_message action with a test `upgrade_notice` value — the `<div class="update-message">` should now render.

(full disclosure: AI helped me identify the issue and verify my work)

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -1095,7 +1095,7 @@ class Antispam_Bee {
 			printf(
 				'<div class="update-message">%s</div>',
 				wp_kses(
-					wpautop( $data['upgrade_notice '] ),
+					wpautop( $data['upgrade_notice'] ),
 					array(
 						'p'     => array(),
 						'a'     => array( 'href', 'title' ),
@@ -3032,7 +3032,7 @@ register_uninstall_hook(
 
 // Upgrade notice.
 add_action(
-	'in_plugin_update_message-' . __FILE__,
+	'in_plugin_update_message-' . plugin_basename( __FILE__ ),
 	array(
 		'Antispam_Bee',
 		'upgrade_notice',


Zodiac1978:
@thisismyurl Thank you so much again for your PR! We sort out, why SonarQube is complaining (looks like a false positive). Otherwise it looks fine for me!

@stklcode Any idea why SonarQube thinks `plugin_basename` has 0 arguments allowed?
Because the spec says it has one: https://developer.wordpress.org/reference/functions/plugin_basename/

stklcode:
> Any idea why SonarQube thinks `plugin_basename` has 0 arguments allowed?

Sonar has no idea about the spec apparently, but it sees the method declaration in the test stubs which indeed has no arguments. One might adjust the stubs here, but the warning has no implications on functionality and may be ignored safely.

https://github.com/pluginkollektiv/antispam-bee/blob/16d4fb31268d0fa53035ddd619c2372d35cfd452/tests/_stubs/includes.php#L33

Marked the issue as "false positive" with a comment.

Zodiac1978 (approved):
Looks fine for me!

2ndkauboy:
@Zodiac1978 we seem to be missing the whole `in_plugin_update_message-{file}` action in `v3`. Should we introduce it back into the new version.

Zodiac1978:
@2ndkauboy I think so, yes. This is a nice feature if we need to show the upgrade notice on the plugin list page also.

2ndkauboy:
I have never used that myself. Is this the feature that displays notes on what changes in the new version on the plugins list? That would be "quite useful" for the update to `v3` I guess. 😅

thisismyurl:
Hi @2ndkauboy and @Zodiac1978 — yes, that's the one. It adds the plugin's `upgrade_notice` readme.txt field as an inline notice below the plugin row when an update is available. Very useful for a v3 major bump since you can call out breaking changes right on the plugins list.

Happy to open a PR against `v3` to bring it back. The implementation would follow the existing module pattern: a small `Admin\UpgradeNotice` class with an `init()` method, registered in `src/load.php` alongside the other modules — that way the `MAIN_PLUGIN_FILE` constant is already in scope without needing a fully qualified reference. Just say the word.

PR #699, opened 2026-06-07
Fix validate_file() check to reject all invalid path codes

### LINKED ISSUE #671: File validation check is too narrow for log file path validation
`validate_file(...) === 1` only checks one invalid case - Safer check would reject all non-zero invalid codes.
https://github.com/pluginkollektiv/antispam-bee/blob/master/antispam_bee.php#L2376

See also: https://developer.wordpress.org/reference/functions/validate_file/

thisismyurl:
validate_file() returns four values:
  0 = valid path
  1 = path contains .. (directory traversal)
  2 = Windows absolute path (drive letter)
  3 = path contains Windows directory separator

The previous check (=== 1) only blocked directory traversal. A Windows-style absolute log path like C:\path\malware.log would return 2 and pass the guard, allowing an unintended write target to be used.

Changing to (!== 0) rejects every invalid code consistently, matching the intent of the original check and the documented behavior of validate_file().

Fixes #671.

### DIFF
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -2394,7 +2394,7 @@ class Antispam_Bee {
 	 * @return  mixed        FALSE in case of error
 	 */
 	private static function _update_spam_log( $comment ) {
-		if ( ! defined( 'ANTISPAM_BEE_LOG_FILE' ) || ! ANTISPAM_BEE_LOG_FILE || ! is_writable( ANTISPAM_BEE_LOG_FILE ) || validate_file( ANTISPAM_BEE_LOG_FILE ) !== 0 ) {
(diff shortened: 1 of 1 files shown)


Zodiac1978:
Hi @thisismyurl, thank you so much for your PR!

I was at Wordcamp Europe in Krakow, sorry for the delay in replying.

Do you want to make the recommended change from CoPilot yourself, so I can merge it? Otherwise, it looks fine.

Zodiac1978 (approved):
Looks fine to me!

thisismyurl:
The reordering is already in this PR. Line 2397 now reads `validate_file( ANTISPAM_BEE_LOG_FILE ) !== 0 || ! is_writable( ANTISPAM_BEE_LOG_FILE )` — `validate_file()` runs first. Copilot appears to have flagged the original code in the diff context. Happy to reword the comment if that would help!

Issue #702, opened 2026-06-11
Implement upgrade_notice for v3


2ndkauboy:
Implement an upgrade notice for version 3 of Antispam Bee to provide users with important information about changes and actions they may need to take following the update. This will help ensure a smooth upgrade experience and proper guidance for existing users.

2ndkauboy:
Merged in https://github.com/pluginkollektiv/antispam-bee/pull/703

PR #703, opened 2026-06-11
feat: restore upgrade notice display to v3

thisismyurl:
## What this does

Restores the `in_plugin_update_message-{file}` upgrade notice that existed in v2 but didn't make it into v3. When WordPress detects an available update for Antispam Bee, the `Upgrade Notice` section from `readme.txt` now appears inline below the plugin row on the plugins list — exactly as it did before.

This is especially useful for the v3 release, where editors will want a heads-up about breaking changes without having to navigate away from the plugins list.

## Changes

**`src/Admin/UpgradeNotice.php`** _(new)_
- Single static class with one public method: `render( array $plugin_data )`
- Guards against an empty notice with `empty()` before outputting anything
- Sanitizes output through `wpautop()` + `wp_kses()` with an explicit allowlist (`p`, `a[href|title]`, `strong`, `em`)

**`src/load.php`**
- Adds `use AntispamBee\Admin\UpgradeNotice` (alphabetically within the Admin group)
- Registers `add_action( 'in_plugin_update_message-' . plugin_basename( MAIN_PLUGIN_FILE ), [ UpgradeNotice::class, 'render' ] )` alongside the other plugin-lifecycle hooks at the bottom of the file, where `MAIN_PLUGIN_FILE` is naturally in scope

## A note on the v2 `wp_kses` attributes

The v2 implementation passed `array( 'href', 'title' )` as the `<a>` attribute list, which is a numeric array and doesn't match the `wp_kses` format (which expects an associative array of `attribute => []`). This PR uses the correct form: `'a' => [ 'href' => [], 'title' => [] ]`.

## Testing

1. Check out the branch and activate the plugin on a local WordPress install.
2. In `readme.txt`, add an `== Upgrade Notice ==` section with a note under any version heading.
3. Temporarily modify the version check so WordPress believes an update is available (or use the [Plugin Update Checker](https://github.com/YahnisElsts/plugin-update-checker) test endpoint).
4. Visit **Plugins → Installed Plugins** — the notice should appear inline below the Antispam Bee row with correct markup.
5. Confirm that an empty `upgrade_notice` field produces no output.

---

Thanks for the great discussion in #698 — happy to tweak anything here.

_This PR was developed with AI assistance (research and implementation). All code was reviewed and validated before submission._

### DIFF
--- a/src/Admin/UpgradeNotice.php
+++ b/src/Admin/UpgradeNotice.php
@@ -13,6 +13,16 @@ namespace AntispamBee\Admin;
 class UpgradeNotice {
 
 	/**
+	 * Register the upgrade notice hook.
+	 */
+	public static function init(): void {
+		add_action(
+			'in_plugin_update_message-' . plugin_basename( \AntispamBee\MAIN_PLUGIN_FILE ),
+			[ __CLASS__, 'render' ]
+		);
+	}
+
+	/**
 	 * Render the upgrade notice below the plugin row on the plugins list.
 	 *
 	 * Called via the `in_plugin_update_message-{file}` action when WordPress
--- a/src/Admin/UpgradeNotice.php
+++ b/src/Admin/UpgradeNotice.php
@@ -18,7 +18,9 @@ class UpgradeNotice {
 	public static function init(): void {
 		add_action(
 			'in_plugin_update_message-' . plugin_basename( \AntispamBee\MAIN_PLUGIN_FILE ),
-			[ __CLASS__, 'render' ]
+			[ __CLASS__, 'render' ],
+			10,
+			2
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


2ndkauboy:
@thisismyurl thanks for this PR! :raised_hands:  I had planned to implement this myself, but you AI bot seem to be have faster. :sweat_smile: 

It only needs to learn how we load our "modules". We don't call hooks in the `load.php` file, but rather have a `init()` function in the modules that would call the hooks.

Which agent are you using BTW? And do you control it or does it just watch for new issue on this repo?

thisismyurl:
Thanks for the kind words, and for the quick turnaround on the feedback, @2ndkauboy! :raised_hands:

I pushed two follow-up commits that move the hook into the module system properly:
- `UpgradeNotice` now has an `init()` method that registers the `in_plugin_update_message-*` action using `\AntispamBee\MAIN_PLUGIN_FILE` (fully qualified, since the class lives in the `Admin` sub-namespace)
- `UpgradeNotice::class` is added to the modules array in `load.php`
- The top-level `add_action()` in `load.php` is removed

As for the AI question — I use [Claude Code](https://claude.ai/code) as part of an internal routing system I've built. It's not an autonomous bot watching for issues on repos. I read the discussion in #698, decided this was worth implementing for the v3 branch, and worked through the code with Claude's help. The system checks my work — reviews output against project conventions, flags issues, and helps me verify the logic before I post anything. I write the direction; it helps me execute and catch mistakes. The module architecture pattern was something I got wrong on the first pass because I was working from `DashboardWidgets` as a reference but missed that lifecycle hooks belong inside the module's `init()` rather than at the top of `load.php`. Glad you caught it.

Happy to address anything else on the implementation.

2ndkauboy:
This closes #702

2ndkauboy:
@thisismyurl thanks for the new changes and your explanation.

I wonder if we should add a `CLAUDE.md`, `AGENTS.md`, etc. files. Or do AI agent also read the `CONTRIBUTING.md` file or can we ask them to in the other files? :thinking:

thisismyurl:
By default, most don't read them but I've setup mine to specifically review the MD files (especially CONTRIBUTE) for rules to follow like which branches to apply to, styles, etc. 

Some repos require an issue to be created for example prior to a PR, some need a specific branch to be targeted, so what I have my "sniffer" do is look at a repo such as yours, have your docs act as gates:

1) read the .MD files, IF QUALIFIED
2) read the issues, IF QUALIFIED
3) read the opened PR's, IF QUALIFIED
4) find branch, IF QUALIFIED
5) send me a note

then, after I've coded a fix

6) rerun steps 1-5, IF QUALIFIED
7) review fix,, IF QUALIFIED
8) post PR

this system cuts out about 90% of potential fixes that I could support, but it helps the "cream" float to the top before I look at it and disqualifies people who've not contributed to their own work for a long time, or don't really want help, or specifically say that AI is unwelcome etc.

short answer, absolutely you should have a contribute as well as funding etc, feel free to look at any of my public projects and borrow from them.

2ndkauboy:
@thisismyurl thanks again for this PR!

I have also created some AI agent files that could be used by other contributers in the future using AI agents: https://github.com/pluginkollektiv/antispam-bee/pull/732

Issue #534, opened 2023-09-22
Feedback from using the new plugin check


Zodiac1978:
There is a new plugin for checking plugins:
https://wordpress.org/plugins/plugin-check/
https://github.com/10up/plugin-check

I have checked ASB with it ang got some issues:

WordPress.Security.NonceVerification.Recommended Line 83 of file inc/columns.class.php.
Processing form data without nonce verification.
`$orderby = isset( $_GET['orderby'] ) ? sanitize_text_field( wp_unslash( $_GET['orderby'] ) ) : '';`

WordPress.Security.NonceVerification.Recommended Line 108 of file inc/columns.class.php.
Processing form data without nonce verification.
`$spam_reason = isset( $_GET['comment_spam_reason'] ) ? sanitize_text_field( wp_unslash( $_GET['comment_spam_reason'] ) ) : '';
`

WordPress.Security.NonceVerification.Recommended Line 133 of file inc/columns.class.php.
Processing form data without nonce verification.
`$spam_reason = isset( $_GET['comment_spam_reason'] ) ? sanitize_text_field( wp_unslash( $_GET['comment_spam_reason'] ) ) : '';
`

WordPress.Security.NonceVerification.Recommended Line 240 of file antispam_bee.php.
Processing form data without nonce verification.
`if ( ! empty( $_GET['comment_status'] ) && 'spam' === $_GET['comment_status'] && ! self::get_option( 'no_notice' ) ) {`

WordPress.Security.NonceVerification.Recommended Line 549 of file antispam_bee.php.
Processing form data without nonce verification.
`return ( ! empty( $_GET['page'] ) && 'antispam_bee' === $_GET['page'] );`

WordPress.WP.AlternativeFunctions.file_system_read_file_put_contents Line 2429 of file antispam_bee.php.
File operations should use WP_Filesystem methods instead of direct PHP filesystem calls. Found: file_put_contents().
`file_put_contents(`

WordPress.WP.AlternativeFunctions.strip_tags_strip_tags Line 2605 of file antispam_bee.php.
strip_tags() is discouraged. Use the more comprehensive wp_strip_all_tags() instead.
`$content = strip_tags( stripslashes( $comment['comment_content'] ) );`

WordPress.WP.AlternativeFunctions.strip_tags_strip_tags Line 2626 of file antispam_bee.php.
strip_tags() is discouraged. Use the more comprehensive wp_strip_all_tags() instead.
`strip_tags( $post->post_title )`

[...]

PR #716, opened 2026-06-17
Escape action link URL and replace strip_tags() with wp_strip_all_tags() (Refs #534)

thisismyurl:
Hey folks,

Two Plugin Check findings from issue #534 that are ready to fix now.

**Settings action link URL not escaped**

`init_action_links()` passes the `add_query_arg()` return value directly into the `href` attribute of the settings link. `add_query_arg()` output should always go through `esc_url()` before it lands in HTML. This fixes the `WordPress.Security.EscapeOutput.OutputNotEscaped` finding.

**`strip_tags()` replaced with `wp_strip_all_tags()` in the notification email**

The spam notification email body uses `strip_tags()` in three places — on `post_title`, `comment_author`, and `comment_content`. `wp_strip_all_tags()` is the WordPress equivalent and handles edge cases (`<script>` and `<style>` content) more completely. The function is available across the supported version range and the output is plain text in all three cases, so the substitution is a safe drop-in.

I left the `file_put_contents` / WP_Filesystem item and the nonce-verification items for a separate PR — those need more context about the surrounding logic to get right.

No unit test added: these are output-escaping and function-preference changes with no behavioral delta, and neither `init_action_links` nor `send_mail_notification` have existing test coverage in the suite.

To verify: activate the plugin on a test site, visit the Plugins page, and confirm the Settings link still resolves correctly. For the `wp_strip_all_tags` change, trigger a spam notification and confirm the email body renders as expected.

(full disclosure: AI helped me identify the issue and verify my work)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 ## Changelog ##
 
+### 2.11.13 ###
+* **English**
+  * Fix: Escape settings action link URL with `esc_url()` (Refs #534) (Thanks @thisismyurl!)
+  * Fix: Replace deprecated `strip_tags()` with `wp_strip_all_tags()` in spam notification email body (Refs #534) (Thanks @thisismyurl!)
+
+* **Deutsch**
+  * Fix: Einstellungs-Link-URL mit `esc_url()` maskieren (Refs #534) (Danke @thisismyurl!)
+  * Fix: Veraltetes `strip_tags()` durch `wp_strip_all_tags()` im Spam-Benachrichtigungs-E-Mail-Body ersetzen (Refs #534) (Danke @thisismyurl!)
+
+
 ### 2.11.12 ###
 * **English**
   * Fix: Fatal error in the dashboard spam counter (Thanks @robertstaddon!)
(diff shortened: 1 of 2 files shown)


Zodiac1978:
@thisismyurl Thank you so much for the PR!

Do you want to modify the changelog according to the CoPilot feedback?

Zodiac1978:
Almost there! I see one newly introduced warning:
https://sonarcloud.io/project/issues?id=pluginkollektiv_antispam-bee&pullRequest=716&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

I don't think this is a major issue, but maybe you can fix this too.

I am still approving this, as the main issue are solved!

Thank you so much for working on those issues :)

Zodiac1978 (approved):
There is one minor warning:
https://sonarcloud.io/project/issues?id=pluginkollektiv_antispam-bee&pullRequest=716&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

Everything else looks fine!

Zodiac1978:
Awesome! Thanks again 🥇

thisismyurl:
Pushed a fix for the SonarCloud warning — split the 164-character `if` condition in `_update_spam_log()` into a multi-line form to bring it under the 160-char limit.

The six other long lines in the file are pre-existing and outside this PR's scope; the SonarCloud `sinceLeakPeriod` filter only flags lines introduced since the last clean point, so they won't block this PR.

Thanks for the kind approval — appreciate you taking the time to review!

Issue #707, opened 2026-06-13
Add Donation, FAQ, Documentation and Support links to general tab


2ndkauboy:
In version 2, we have these four links

<img width="372" height="141" alt="Image" src="https://github.com/user-attachments/assets/15ba0da2-c61b-4e70-9547-24dfe1654c3d" />

In version 3, they are missing. In #461 we updated some of them. Do we want to reintroduce those four links on the "General" tab?

krafit:
Yes we should reintroduce them, @2ndkauboy

2ndkauboy:
Closed with https://github.com/pluginkollektiv/antispam-bee/pull/724

PR #720, opened 2026-06-26
feat: add Support tab to settings page

2ndkauboy:
## Summary

- Adds a **Support** tab after General, Comment and Linkback, restoring the resource links that existed in v2 as a sidebar
- The tab contains four sections: **Donate**, **FAQ**, **Documentation** and **Support**, each with a short description and a button link
- The settings form and save button are hidden when the Support tab is active, since it holds no configurable options

### DIFF
--- a/src/Admin/SettingsPage.php
+++ b/src/Admin/SettingsPage.php
@@ -116,6 +116,9 @@ class SettingsPage {
 				ContentTypeHelper::get_type_name( $type )
 			);
 		}
+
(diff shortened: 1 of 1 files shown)


2ndkauboy:
This closes #707

2ndkauboy:
This was the proposed design:
image

It decided against this quite large tab and to use simple links on the "General" tab instead, as implemented in #724.

PR #729, opened 2026-06-28
fix: only determine client IP from REMOTE_ADDR via pre_comment_user_ip filter

### LINKED ISSUE #708: Port #513 into v3
The changes user IP resolution from #513 was not ported into `v3`.

2ndkauboy:
## Summary

- Refactors `IpHelper::get_client_ip()` to use only `REMOTE_ADDR` by default instead of checking multiple proxy headers (which are trivially spoofable).
- Proxy/forwarded headers can be used by hooking the standard WordPress `pre_comment_user_ip` filter.
- Updates `CommentTest` to use `REMOTE_ADDR` instead of `HTTP_CLIENT_IP`.
- Adds a new `tests/Unit/Helpers/IpHelperTest.php` with tests for default behaviour and filter override.

Closes #708

## Test plan

- [ ] Run `composer test:unit` — all tests pass
- [ ] Verify client IP is read from `REMOTE_ADDR` by default
- [ ] Verify a `pre_comment_user_ip` filter hook can supply an IP from a trusted proxy header

### DIFF
--- a/src/Helpers/IpHelper.php
+++ b/src/Helpers/IpHelper.php
@@ -13,38 +13,25 @@ namespace AntispamBee\Helpers;
 class IpHelper {
 
 	/**
-	 * Return real client IP
+	 * Return real client IP.
+	 *
+	 * By default, only `REMOTE_ADDR` is evaluated. Use the `pre_comment_user_ip`
+	 * filter to supply an IP from a trusted proxy header instead.
+	 *
+	 * @hook    string  pre_comment_user_ip  The client IP, defaults to REMOTE_ADDR.
 	 *
 	 * @return string Client IP
 	 */
 	public static function get_client_ip(): string {
+		// phpcs:disable WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
 		// phpcs:disable WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
-		// Sanitization of $ip takes place further down.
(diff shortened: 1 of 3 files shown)


src/Helpers/IpHelper.php
```diff
@@ -13,38 +13,25 @@
...
-		if ( $ip ) {
-			return $ip;
-		}
-
-		if ( isset( $_SERVER['REMOTE_ADDR'] ) ) {
-			return self::sanitize_ip( wp_unslash( $_SERVER['REMOTE_ADDR'] ) );
-		}
-
-		return '';
+		// phpcs:disable WordPress.Security.ValidatedSanitizedInput.InputNotValidated
```
stklcode, line 28: Not really elegant either, but at least only one (long) line instead of 6...

```php
		return self::sanitize_ip(
			// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.ValidatedSanitizedInput.InputNotValidated
			(string) apply_filters( 'pre_comment_user_ip', wp_unslash( $_SERVER['REMOTE_ADDR'] ?? '' ) )
		);
```

2ndkauboy: I'd rather have shorter lines, like in the original code. But yes, this is uggly!

PR #731, opened 2026-06-28
fix: make quotes optional in Honeypot inject() regex; add tests

### LINKED ISSUE #713: Port #658 to v3
The code changes are already in #676, but the new tests from #658 are missing.

2ndkauboy:
## Summary

- Makes attribute quotes optional in the `Helpers/Honeypot::inject()` regex patterns (e.g. `id=["\']?comment["\']?`), so it handles HTML with unquoted attributes.
- Adds a new `tests/Unit/Helpers/HoneypotHelperTest.php` covering:
  - `inject()` returns `''` when the target field id is not found
  - `inject()` works with quoted `name`/`id` attributes (baseline)
  - `inject()` works with unquoted `name`/`id` attributes

The code fix was supposed to land via #676 but the regex in `src/Helpers/Honeypot.php` was still requiring quotes.

Closes #713

## Test plan

- [ ] Run `composer test:unit` — all 15 tests pass including the 3 new Honeypot helper tests
- [ ] Manually verify a comment form with unquoted textarea attributes is handled correctly

### DIFF
--- a/src/Helpers/Honeypot.php
+++ b/src/Helpers/Honeypot.php
@@ -81,17 +81,17 @@ class Honeypot {
 						<textarea                                        (?# the opening of the textarea and some optional attributes )
 						(                                                (?# match a id attribute followed by some optional ones and the name attribute )
 							(?P<before1>[^>]*)
-							(?P<id1>id=["\']{{HONEYPOT_ID}}["\'])
+							(?P<id1>id=["\']?{{HONEYPOT_ID}}["\']?)
 							(?P<between1>[^>]*)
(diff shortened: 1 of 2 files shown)


stklcode:
> The code fix was supposed to land via #676 but the regex in `src/Helpers/Honeypot.php` was still requiring quotes.

Actually, the proposed code change corresponds exactly to the one in #676 - plus tests, so let's prefer this one.

I would add some reference to the original issue (#657) or the v2 change (4e36755df73cba0942df28d761d7ee447f12b9d3, #658) and a word or two about the actual issue in the commit message, but that's just personal preference.

PR #732, opened 2026-06-28
chore: add AI agent contribution guidelines

2ndkauboy:
## Summary

- Add `AGENTS.md` as the single source of truth for AI agent contribution rules: code quality checks, git workflow, conventional commits, PR format, and ignore/export file conventions
- Add tool-specific stub files that each point to `AGENTS.md` so rules only need to be maintained in one place
- Exclude all added files from plugin distributions via `.distignore` and `git archive` via `.gitattributes`
- Extend `.gitignore` with local AI agent config files (`.claude/`, `.aider/`, `*.local.md`, etc.)
- Normalise `.gitignore`, `.distignore`, and `.gitattributes`: consistent `/` prefixes, remove duplicate directory entries, document and apply the correct conventions for each file

| File | AI agent |
|------|----------|
| `AGENTS.md` | OpenAI Codex (native), any agent that reads it |
| `CLAUDE.md` | Claude Code |
| `.github/copilot-instructions.md` | GitHub Copilot |
| `.cursor/rules/agents.mdc` | Cursor |
| `.windsurfrules` | Windsurf |
| `CONVENTIONS.md` | Aider |

## Test plan

No functional changes — documentation/config files only. No test steps required.


stklcode (changes requested):
All new files/directories should be added to _.distignore_ and probably _.gitattributes_ if they don't belong to the final package.

2ndkauboy:
@stklcode that's correct. I'll add them.

AGENTS.md
```diff
@@ -0,0 +1,56 @@
+# AI Agent Guidelines
+
+General rules for AI agents contributing to this project.
+
+## Code quality — verify before every commit
+
+| Check                   | Command              |
+|-------------------------|----------------------|
+| PHP code style          | `composer cs`        |
+| PHPStan static analysis | `composer analyse`   |
```
thisismyurl: `composer analyse` isn't defined in `composer.json` yet — the available scripts are `cs`, `csfix`, `lint-php`, and `test:unit` (PHPStan looks like it's still on the `feat/phpstan` branch). Worth either adding an `analyse` script or dropping this row until it lands, so an agent's pre-commit checklist doesn't error on this step.

AGENTS.md
```diff
@@ -0,0 +1,56 @@
...
+| PHP unit tests          | `composer test:unit` |
+| E2E tests               | `npm run test:e2e`   |
```
thisismyurl: `npm run test:e2e` (`wp-scripts test-playwright`) needs the wp-env environment running first, or it fails cold.

```suggestion
| E2E tests               | `npm run env:start`, then `npm run test:e2e` |
```

AGENTS.md
```diff
@@ -0,0 +1,56 @@
...
+
+Fix code style violations automatically with `composer csfix`.
+
+## Git workflow
+
+- Always work on a **feature branch** based on `develop`
+- Never commit directly to `develop` or `master`
```
thisismyurl: There's no `develop` branch in the repo — the default is `master` and the active integration branch looks like `v3` (this PR's own base). Adjusting to match, but you'll know the intended model best.

```suggestion
- Always work on a **feature branch** based on `v3`
- Never commit directly to `v3` or `master`
```

AGENTS.md
```diff
@@ -0,0 +1,56 @@
...
+Excludes tracked files from `git archive` exports via `export-ignore`.
+
+- Only tracked files and directories need an entry here — gitignored paths are already absent from archives
+- Always use `/` prefix (e.g. `/.github export-ignore`)
+- No trailing `/` on directories (not needed)
+- Keep each section (`# Directories`, `# Files`) sorted alphabetically
+
+## Pull requests
+
+- Base branch: `develop`
```
thisismyurl: Same as above — this PR targets `v3`, not `develop`.

```suggestion
- Base branch: `v3`
```

.gitignore
```diff
@@ -1,20 +1,29 @@
...
+/.wp-env.override.json
+/AGENTS.local.md
+/CLAUDE.local.md
+/CONVENTIONS.local.md
+/DOCKER_ENV
+/chromedriver
+/composer.lock
+/driver.tar.gz
+/driver.zip
+/geckodrier
```
thisismyurl: Small typo: the Mozilla WebDriver binary is `geckodriver`; `geckodrier` won't match it.

```suggestion
/geckodriver
```

.distignore
```diff
@@ -1,31 +1,53 @@
...
 /behat.yml
+/chromedriver
 /composer-php-5-4.json
 /composer.json
 /composer.lock
 /docker-compose.yml
 /docker_tag
+/driver.tar.gz
+/driver.zip
+/geckodrier
```
thisismyurl: Same typo here — `geckodriver`.

```suggestion
/geckodriver
```

thisismyurl (commented):
Thanks for putting this together — the single-source-of-truth `AGENTS.md` with thin per-tool stubs is a clean pattern, and the ignore / export-file conventions are genuinely well documented. I ran the referenced commands and branch names against the repo and a few don't match the current setup yet, so I've left inline suggestions for the unambiguous ones and notes for the judgment calls. All entirely your call.

**Minimum PHP version.** `composer.json` requires `php >=7.2`, and AI agents tend to reach for 7.4/8.x syntax by default. A one-liner in the *Code quality* section would save a round of CS failures, e.g.:
> - Target **PHP 7.2+** — avoid syntax introduced in 7.4 or later (typed properties, arrow functions, `match`, null-coalescing assignment `??=`).

(The leftover `composer-php-5-4.json` might be worth a word too, so an agent doesn't mistake 5.4 for the floor.)

**Security conventions.** Since this is a spam-filtering plugin, a short *WordPress & security* section — nonces, capability checks, sanitisation/escaping, the text domain — would probably do more for contribution quality than anything else here. Optional, but high-value.

The inline notes cover the `develop`-vs-`v3` branch references, the `composer analyse` script, the e2e environment step, and a couple of typos. Thanks again — happy to help with any of these.

AGENTS.md
2ndkauboy: Since I plan to merge the PHPStan branch before this in, I already added it, so I don't forget it. 😅

AGENTS.md
2ndkauboy: My plan is to rename `v3` to `develop` and `master` to `v2` or `legacy`. But maybe we should have the current branches here and update them after we released `v3`.

2ndkauboy:
@thisismyurl thanks for all your recommendations. I've added them to the files and adjusted some other small things.

.gitattributes
```diff
@@ -1,4 +1,6 @@
 # Directories
+/.cursor export-ignore
+/.git export-ignore
```
stklcode: redundant, .git (built-in) is handled properly by default
```suggestion
```

.gitattributes
```diff
@@ -11,15 +13,23 @@
 /.eslintrc.json export-ignore
 /.gitattributes export-ignore
 /.gitignore export-ignore
+/.phpunit.result.cache export-ignore
```
stklcode: Quoting the new guideline :wink:
> Only tracked files and directories need an entry here ...

`git export` does not care about untracked files. This is rather a candidate for .gitignore than for .gitattributes

Same applies to some other entries that do not exist in the actual source tree.

antispam_bee.php
```diff
@@ -24,22 +24,24 @@
 define( __NAMESPACE__ . '\MAIN_PLUGIN_FILE', __FILE__ );
 define( __NAMESPACE__ . '\PLUGIN_PATH', plugin_dir_path( MAIN_PLUGIN_FILE ) );
 
-// The pre_init functions check the compatibility of the plugin and calls the init function, if check were successful.
+// The pre_init functions check the compatibility of the plugin and calls the init function if check were successful.
```
stklcode: Changes to this look good, but are completely unrelated to the proposal of adding contribution guidelines. I’d rather split these out or at least keep them in a separate commit.

composer.json
```diff
@@ -27,7 +27,6 @@
   "require-dev": {
     "dealerdirect/phpcodesniffer-composer-installer": "^v1.0.0",
     "phpcompatibility/phpcompatibility-wp": "^2.1",
-    "wp-coding-standards/wpcs": "^3.0.1",
```
stklcode, line 30: OK (remove duplicate dependency declaration), but completely unrelated to both PR and commit message.

antispam_bee.php
2ndkauboy: Those should not be in this PR, you are right! 🙈

.gitattributes
2ndkauboy: Correct, I have updated those.

PR #733, opened 2026-06-28
ci: simplify test matrix and update actions

stklcode:
* eliminate the `core` variable which can be trivially derived from `wordpress` version
* update `actions/checkout` to v7 and use same version for all workflows
* update `actions/setup-node` to v6
* update `actions/upload-artifact` to v7
* update Node.js to v22


2ndkauboy:
Cool! Didn't know we don't need the ZIP file paths. Even better!

PR #759, opened 2026-07-03
ci: add spell-checking workflow

### LINKED ISSUE #704: Add more GitHub Actions (Typos, PHPStan, Plugin Check)
**Is your feature request related to a problem? Please describe.**
No, just more checks.

**Describe the solution you'd like**
Here is an example from my plugin:
https://github.com/Zodiac1978/plugin-report/blob/develop/.github/workflows/ci.yml

**Describe alternatives you've considered**
We already have some checks, but none for typos or the "plugin check" plugin checks ;)

2ndkauboy:
Closes #704

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -24,3 +24,4 @@
 /phpcs.xml
 /phpunit.xml.dist
 /playwright.config.ts
+/typos.toml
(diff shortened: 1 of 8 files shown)


2ndkauboy:
@Zodiac1978 the spell checker marks every German entry in the `CHANGELOG.md` file. What should we do about it? Ignore the file or remove the German translations? We could move them to a `CHANGELOG.de.md` or just drop them altogether. Is this file even used on WordPress.org or does it only read the changelog section from the readme?

2ndkauboy:
There seems to be a config file I can ignore these lines from.

PR #761, opened 2026-07-06
fix: mask IP addresses to a `/24` (IPv4) and `/48` (IPv6) network

2ndkauboy:
## Description

`\AntispamBee\Helpers\IpHelper::anonymize_ip` slices an IP address with the pattern `\w+([\.:])\w+` and keeps what it matched: the first two octets of an IPv4 address (a `/16`) and the first two groups of an IPv6 address (a `/32`).

This masks the address instead, with a **`/24`** for IPv4 — the mask WordPress core uses in `wp_privacy_anonymize_ip()` — and a **`/48`** for IPv6:

```php
public static function anonymize_ip( string $ip ): string {
	$packed = inet_pton( $ip );

	if ( false === $packed ) {
		return '';
	}

	// IPv4 addresses pack into four bytes, IPv6 addresses into sixteen.
	if ( 4 === strlen( $packed ) ) {
		$netmask = '255.255.255.0';
	} elseif ( 0 === strncmp( $packed, str_repeat( "\0", 10 ) . "\xff\xff", 12 ) ) {
		// An IPv4-mapped IPv6 address carries an IPv4 address, mask that one.
		$netmask = '::ffff:255.255.255.0';
	} else {
		$netmask = 'ffff:ffff:ffff::';
	}

	$anonymized = inet_ntop( $packed & (string) inet_pton( $netmask ) );

	return false === $anonymized ? '' : $anonymized;
}
```

Input that cannot be parsed as an IP address results in an empty string, so `CountrySpam` never sends anything but an address to the geolocation service. `inet_pton` returns `false` for such input without emitting a warning, so it doubles as the validation guard.

## Behaviour

| Input | Before | After | WP core |
|---|---|---|---|
| `198.51.100.42` | `198.51.0.0` | `198.51.100.0` | `198.51.100.0` |
| `10.11.12.13` | `10.11.0.0` | `10.11.12.0` | `10.11.12.0` |
| `2001:db8:85a3:8d3:1319:8a2e:370:7348` | `2001:db8::` | `2001:db8:85a3::` | `2001:db8:85a3:8d3::` |
| `2001:db8:85a3::8a2e:370:7334` | `2001:db8::` | `2001:db8:85a3::` | `2001:db8:85a3::` |
| `2a02:8109:aa00:1234::5` | `2a02:8109::` | `2a02:8109:aa00::` | `2a02:8109:aa00:1234::` |
| `2001:db8::1` | `2001:db8::` | `2001:db8::` | `2001:db8::` |
| `::1` | `''` | `::` | `::` |
| `0:0:0:0:0:0:0:1` | `0:0::` | `::` | `::` |
| `fe80::1` | `''` | `fe80::` | `fe80::` |
| `::ffff:192.0.2.123` | `ffff:192::` | `::ffff:192.0.2.0` | `::ffff:192.0.2.0` |
| `''` | `''` | `''` | `0.0.0.0` |
| `no-ip-at-all` | `''` | `''` | `0.0.0.0` |

➡️ IPv4 is now identical to core's. IPv6 is deliberately one step **coarser** than core, and the result deliberately differs for input that cannot be parsed: core answers `0.0.0.0`, which would be looked up as if it were a real address, so an empty string is kept instead.

Besides the masks, this fixes three defects of the pattern:

- A valid address it could not match was **rejected outright** — `::1` and `fe80::1` came back as an empty string.
- The result depended on the **notation** rather than the address: `::1` was rejected, while the very same address written `0:0:0:0:0:0:0:1` became `0:0::`.
- An **IPv4-mapped IPv6 address** produced an unrelated prefix, turning `::ffff:192.0.2.123` into `ffff:192::`. Such an address can show up in `REMOTE_ADDR` on a dual-stack host, and `filter_var( …, FILTER_VALIDATE_IP )` in `get_client_ip()` accepts it, so it does reach this method. It is now masked like the IPv4 address it carries, as core does.

## Why `/24` for IPv4

Geolocation databases key the country on `/24`, and a single `/16` can span country borders, so the previous mask could resolve to the wrong country or to no country at all. Neither failure is visible to the site owner:

- `CountrySpam::verify` returns `0` when the response carries no `country_code`, so an unresolvable network address silently means "not spam".
- In allow-list mode, a neighbouring country that is not on the list turns a **legitimate** comment into spam.

A `/24` is also the boundary core and the common data-protection guidance have settled on, and it is what Antispam Bee 2.x shipped.

## Why `/48` for IPv6

2.x kept only the **first group** of an IPv6 address. That looks like a stronger privacy guarantee, but a `/16` cannot identify a country at all, so what it really did was disable the rule for IPv6 visitors: RIPE sub-allocates `2a00::/12` in `/29`–`/32` chunks, so three different countries collapsed onto one identical query and anything starting with `2001:` collapsed onto the Teredo prefix.

The mask has to be coarse enough to drop the host and fine enough to resolve a country. `/48` is the coarsest mask that satisfies both, and this was measured rather than assumed — see [the detailed comment below](https://github.com/pluginkollektiv/antispam-bee/pull/761#issuecomment-5251105039). Against 1592 IPv6 addresses taken from real comments:

- **No announced BGP prefix is longer than `/48`** — 0 of 1388 addresses with a route. Every `/64` inside a `/48` therefore falls in the same announced prefix, so a `/48` mask cannot change which block a lookup matches. Of the 125 `/48` groups holding more than one distinct `/64`, none has a conflicting country.
- Sampling 225 networks against IPLocate at four widths, the country was **identical at `/64`, `/56` and `/48` for all 225**. `/40` is the first mask that changes the answer, flipping 13 of them — tunnel brokers and hosters where the geo layer localises below the announced `/32`, so the lookup returns the provider's country instead of the visitor's.

A `/64` — what core uses — is a single subscriber LAN: ISPs delegate one per VLAN, and cellular links and hosted servers commonly get exactly one. `/48` is the site allocation, so it drops the subscriber without costing a single country lookup. Bit symmetry with IPv4 is not a useful yardstick either: an IPv4 `/16` is up to 65,536 addresses across many organisations, while an IPv6 `/64` is one LAN.

Worth stating plainly: `/48` sits **exactly** on the boundary, with 66 of the measured networks announced at precisely `/48`. That is an argument for recording the measurement, not for backing off — anything a vendor keys finer than `/48` would be city-level data, which this rule does not use.

## Why not call `wp_privacy_anonymize_ip()` directly

It requires WordPress 4.9.6, while we still declare `Requires at least: 4.7`. Delegating to it with a bundled fallback was explored in #762, where the two code paths ended up applying different masks. It would also fix the IPv6 mask at core's `/64` and answer `0.0.0.0` for unparseable input. Core's extra handling of ports, brackets and zone identifiers is not needed here, because `get_client_ip()` has already put the address through `filter_var( …, FILTER_VALIDATE_IP )`.

Core's function also exists for a different purpose — storing a visitor's IP safely *inside* the site — whereas this value is sent to a third party on every reaction, which is what justifies going one step coarser.

## Testing

Extends the existing `anonymize_ip_provider` in `IpHelperTest` to cover every row of the table above. The IPv4 rows deliberately use addresses whose second and third octets are non-zero, and `2a02:8109:aa00:1234::5` is there so that an implementation keeping a `/56` — the fourth group truncated rather than dropped — would fail. Every expectation was cross-checked against an independent implementation.

`composer test:unit` — 69 tests pass. `phpcs` and `phpstan` (level 8) clean.

## Out of scope

`CountrySpam` sends the anonymized IP to iplocate.io for every reaction, including loopback and private addresses, and interpolates the result into the URL without checking for an empty string. Worth a separate issue.

### DIFF
--- a/tests/Unit/Helpers/IpHelperTest.php
+++ b/tests/Unit/Helpers/IpHelperTest.php
@@ -55,8 +55,9 @@ class IpHelperTest extends TestCase {
 		return [
 			[ '198.51.100.42', '198.51.100.0', 'IPv4 address should be truncated to a /24 network' ],
 			[ '10.11.12.13', '10.11.12.0', 'IPv4 host bits should be zeroed out' ],
-			[ '2001:db8:85a3:8d3:1319:8a2e:370:7348', '2001:db8:85a3:8d3::', 'IPv6 address should be truncated to a /64 network' ],
-			[ '2001:db8:85a3::8a2e:370:7334', '2001:db8:85a3::', 'compressed IPv6 address should be truncated to a /64 network' ],
+			[ '2001:db8:85a3:8d3:1319:8a2e:370:7348', '2001:db8:85a3::', 'IPv6 address should be truncated to a /48 network' ],
+			[ '2001:db8:85a3::8a2e:370:7334', '2001:db8:85a3::', 'compressed IPv6 address should be truncated to a /48 network' ],
+			[ '2a02:8109:aa00:1234::5', '2a02:8109:aa00::', 'the fourth group should be dropped whole, not truncated to a /56' ],
 			[ '2001:db8::1', '2001:db8::', 'IPv6 address shorter than the mask should only lose its host bits' ],
 			[ '::1', '::', 'the IPv6 loopback address should be masked, not rejected' ],
 			[ '0:0:0:0:0:0:0:1', '::', 'the result should not depend on the notation of the address' ],
--- a/tests/Unit/Helpers/IpHelperTest.php
+++ b/tests/Unit/Helpers/IpHelperTest.php
@@ -57,7 +57,7 @@ class IpHelperTest extends TestCase {
 			[ '10.11.12.13', '10.11.12.0', 'IPv4 host bits should be zeroed out' ],
 			[ '2001:db8:85a3:8d3:1319:8a2e:370:7348', '2001:db8:85a3::', 'IPv6 address should be truncated to a /48 network' ],
 			[ '2001:db8:85a3::8a2e:370:7334', '2001:db8:85a3::', 'compressed IPv6 address should be truncated to a /48 network' ],
-			[ '2a02:8109:aa00:1234::5', '2a02:8109:aa00::', 'the fourth group should be dropped whole, not truncated to a /56' ],
+			[ 'fd02:8109:aa00:1234::5', 'fd02:8109:aa00::', 'the fourth group should be dropped whole, not truncated to a /56' ],
 			[ '2001:db8::1', '2001:db8::', 'IPv6 address shorter than the mask should only lose its host bits' ],
 			[ '::1', '::', 'the IPv6 loopback address should be masked, not rejected' ],
 			[ '0:0:0:0:0:0:0:1', '::', 'the result should not depend on the notation of the address' ],
--- a/tests/Unit/Helpers/IpHelperTest.php
+++ b/tests/Unit/Helpers/IpHelperTest.php
@@ -53,10 +53,16 @@ class IpHelperTest extends TestCase {
 
 	public function anonymize_ip_provider(): array {
 		return [
-			[ '192.0.2.123', '192.0.0.0', 'IPv4 address should be truncated' ],
-			[ '2001:db8:85a3::8a2e:370:7334', '2001:db8::', 'IPv6 address should be truncated' ],
+			[ '198.51.100.42', '198.51.100.0', 'IPv4 address should be truncated to a /24 network' ],
+			[ '10.11.12.13', '10.11.12.0', 'IPv4 host bits should be zeroed out' ],
+			[ '2001:db8:85a3:8d3:1319:8a2e:370:7348', '2001:db8:85a3:8d3::', 'IPv6 address should be truncated to a /64 network' ],
+			[ '2001:db8:85a3::8a2e:370:7334', '2001:db8:85a3::', 'compressed IPv6 address should be truncated to a /64 network' ],
+			[ '2001:db8::1', '2001:db8::', 'IPv6 address shorter than the mask should only lose its host bits' ],
+			[ '::1', '::', 'the IPv6 loopback address should be masked, not rejected' ],
+			[ '0:0:0:0:0:0:0:1', '::', 'the result should not depend on the notation of the address' ],
+			[ 'fe80::1', 'fe80::', 'a link-local IPv6 address should be masked, not rejected' ],
+			[ '::ffff:192.0.2.123', '::ffff:192.0.2.0', 'an IPv4-mapped IPv6 address should be masked like an IPv4 address' ],
 			[ '', '', 'empty input should result in an empty string' ],
-			[ '::1', '', 'unmatchable input should result in an empty string' ],
 			[ 'no-ip-at-all', '', 'non-IP input should result in an empty string' ],
 		];
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
       its words with spaces, for example Chinese, Japanese, Korean or Thai
     * Fix: The language rule no longer marks a comment as spam if the language could not be
       determined at all
+    * Fix: IP addresses are now anonymized by masking the host portion of the address, using the
+      same networks as WordPress itself: a /24 for IPv4, as in Antispam Bee 2.x, and a /64 for
+      IPv6 instead of only its first two groups
 
 * **Deutsch**
     * Kompletter Code-Rewrite und Überarbeitung des Backend-User-Interfaces
@@ -20,6 +23,9 @@
       durch Leerzeichen trennt, beispielsweise Chinesisch, Japanisch, Koreanisch oder Thai
     * Fix: Die Sprach-Regel markiert einen Kommentar nicht mehr als Spam, wenn die Sprache
       überhaupt nicht bestimmt werden konnte
+    * Fix: IP-Adressen werden nun anonymisiert, indem der Host-Teil der Adresse maskiert wird, und
+      zwar mit denselben Netzen wie in WordPress selbst: ein /24 für IPv4, wie in Antispam Bee 2.x,
+      und ein /64 für IPv6 statt nur der ersten zwei Blöcke
 
 ### 2.11.12 ###
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,9 @@
       its words with spaces, for example Chinese, Japanese, Korean or Thai
     * Fix: The language rule no longer marks a comment as spam if the language could not be
       determined at all
-    * Fix: IP addresses are now anonymized by masking the host portion of the address, using the
-      same networks as WordPress itself: a /24 for IPv4, as in Antispam Bee 2.x, and a /64 for
-      IPv6 instead of only its first two groups
+    * Fix: IP addresses are now anonymized by masking the host portion of the address: a /24 for
+      IPv4, the same network WordPress itself keeps, as in Antispam Bee 2.x, and a /48 for IPv6
+      instead of only its first two groups
 
 * **Deutsch**
     * Kompletter Code-Rewrite und Überarbeitung des Backend-User-Interfaces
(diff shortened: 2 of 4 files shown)


stklcode (commented):
I like the new implementation as well as the tests :+1: 

> [...] while an IPv6 `/64` is a single subscriber LAN. The mask has to be coarse enough to drop the host and fine enough to resolve a country — the trade-off core already made.

Well ... depending on your definition of "single subscriber".

My uplinks have a `/48` (business plan) and a `/56` prefix (backup consumer plan) assigned from my ISPs - as recommended by RIPE[^1] - so in the final setup `/64` is delegated down to each VLAN. Cellular links typically get `/64` prefixes per UE, as well as servers in typical hosting environments.

Think I would have aimed for `/56` for "site" granularity which should be more than precise enough for country detection and pretty safe to send to external lookup APIs. Can't see no reason to be more precise than that for our purpose.

(~~btw. that's what WP core does, I guess for about the same resaons~~ wrong, see below)

[^1]: https://www.ripe.net/publications/docs/ripe-690/ (§4.2.2)

2ndkauboy:
> (btw. that's what WP core does, I guess for about the same resaons)

Are you sure about that? I have to admit that I have not tested this myself with all the examples, but Claude said that it uses /64 (as written in the table).

stklcode:
Right, looks like I confused myself with a false statement from #762, `/64` is correct for the core function:
```php
		$netmask       = 'ffff:ffff:ffff:ffff:0000:0000:0000:0000';
/* ... */
			$ip_addr = inet_ntop( inet_pton( $ip_addr ) & inet_pton( $netmask ) );
```
https://github.com/WordPress/wordpress-develop/blob/7.0.3/src/wp-includes/functions.php#L8316

----

I would still prefer `/56` for this use case for the reasons mentioned above, but matching a core function is also a valid point.

2ndkauboy:
I measured the `/56` vs `/64` question against real traffic instead of arguing it, and the answer turns out to be that we can go further than either: **`/48` costs nothing.**

Source data is a comments dump: 1592 valid global IPv6 addresses, 1491 distinct `/64`s in 643 distinct `/48`s. Three independent lines of evidence.

### 1. Structural — this one is a proof, not a sample

Queried Team Cymru for the BGP-announced prefix behind every one of the 1491 distinct `/64`s (1388 have a route; 103 have none and resolve to no country at any mask width):

| longest announced prefix | count |
|---|---|
| **longer than `/48`** | **0** |
| exactly `/48` | 66 |
| `/29` (most common) | 564 |
| `/35` | 325 |
| `/32` | 177 |

Nothing is announced longer than `/48`. Every `/64` inside a `/48` therefore falls in the same announced prefix, so masking to `/48` **cannot** change which block a lookup matches. Confirmed independently: 125 of the `/48` groups contain more than one distinct `/64`, and **zero** have conflicting countries. The first genuine conflict appears only at `/32` (`2001:67c::` mixes RO and SE — shared PI space).

### 2. What each mask would actually break

| mask | networks pulled outside their announced prefix |
|---|---|
| `/64` | 0 / 1388 (0.0%) |
| `/56` | 0 / 1388 (0.0%) |
| `/48` | 0 / 1388 (0.0%) |
| `/40` | 146 / 1388 (**10.5%**) |
| `/32` | 530 / 1388 (38.2%) |

### 3. Empirical — against iplocate itself

225 networks sampled at random (seeded) from the 1491, each looked up at four widths, 556 lookups with a real API key:

| mask | same country as `/64` | differs | agreement |
|---|---|---|---|
| `/56` | 225 | 0 | **100.00%** |
| `/48` | 225 | 0 | **100.00%** |
| `/40` | 212 | 13 | 94.22% |

Zero divergence at `/48`. The 13 `/40` failures are real country flips, not lookup failures, grouped by the announced route (individual visitor prefixes omitted deliberately):

| announced route | flip | count |
|---|---|---|
| `2a09:bac5::/32` | TR → TH | 3 |
| `2001:470::/32` (HE tunnel broker) | GB → US | 2 |
| `2001:470::/32` | US → BR | 2 |
| `2607:740:24::/48` | EG → US | 1 |
| `2607:740:3d::/48` | MX → US | 1 |
| `2001:ac8:20::/48` | DE → GB | 1 |
| `2001:67c:187c::/48` | RO → PL | 1 |
| `2607:f7a0:f::/48` | CA → US | 1 |
| (no route) | BN → GB | 1 |

These are tunnel brokers and hosters where the geo layer localises well below the announced `/32`, so a `/40` mask returns the provider's home country instead of the visitor's. In allow-list mode that turns a legitimate comment from Egypt or Mexico into spam — exactly the failure this PR is trying to remove.

### So

`/48` is the coarsest mask that provably loses nothing: three steps more private than the `/64` this PR currently implements, one more than the `/56` suggested above, and the first measurable cost only appears at `/40`. Only the netmask changes:

```php
$netmask = 'ffff:ffff:ffff::';
```

Two caveats, stated plainly:

- `/48` sits **exactly** on the boundary with zero margin — 66 networks are announced at precisely `/48`. That is an argument for documenting the measurement, not for backing off to `/56`; anything a vendor keys finer than `/48` would be city-level data, which this rule does not use.
- The iplocate sample is 225 networks (95% upper bound on the `/48` error rate ≈ 1.3%), limited by the API's 1000 lookups/day. The structural result in (1) is the stronger evidence and covers the full population.

On the point about matching core: we already deliberately diverge from `wp_privacy_anonymize_ip()` on unparseable input (`''` rather than `0.0.0.0`), so the property was never "identical to core" but "identical to core for parseable addresses". Relaxing that to "at least as coarse as core" is a table row and a docblock sentence. Core's function also exists for GDPR-safe logging of an IP *inside* the site, which is a different threat model from shipping the value to a third-party API on every reaction.

<sub>Measurements: Team Cymru bulk WHOIS for announced prefixes, iplocate.io for countries. Full addresses never left the machine — only `/64`-masked forms went to Cymru, and `/64`…`/40`-masked forms to iplocate.</sub>

2ndkauboy:
@stklcode what do you think of that analysis? Use `/48` or be a little safer with `/56`?

stklcode:
That’s mor or less what I would have expected 😄 

From both recommendations and real-world data I’d say it’s pretty much equivalent for our use case.
`/56` is slightly more focused when it comes down to blocking/allowing specific network segments. For rough origin-detection every typical assignment smaller than `/32` (or `/40` from the analysis) should be rather OK.

`/29` - typical LIR assignment
`/32` - typical ISP assignment
`/48` - recommended enterprise/site assignment
`/56` - recommended residential assignment
`/60` - small residential or sub-network assignment
`/64` - typical LAN segment

2ndkauboy:
So do we use `/48` or `/56` for now with iplocate?

stklcode:
Are there any plans to re-use the resulting IPs for other rules?

`/48` should be the privacy-first choice here and good enough for the given purpose.

Otherwise if we consider IP-based filter rules I’d go for `/56`. But I guess that can be changed in future versions without much loss if we go for 48 for now.

2ndkauboy:
This is only for the request to iplocate.io for now. Other things should use the unanonymized IP address, like the spam log entry, so fail2ban can act on it. Putting the anonymized IP into the logfile would mean blocking a whole country. :see_no_evil:

tests/Unit/Helpers/IpHelperTest.php
```diff
@@ -53,10 +53,17 @@ public function test_anonymize_ip( string $ip, string $expected, string $message
 
 	public function anonymize_ip_provider(): array {
 		return [
-			[ '192.0.2.123', '192.0.0.0', 'IPv4 address should be truncated' ],
-			[ '2001:db8:85a3::8a2e:370:7334', '2001:db8::', 'IPv6 address should be truncated' ],
+			[ '198.51.100.42', '198.51.100.0', 'IPv4 address should be truncated to a /24 network' ],
+			[ '10.11.12.13', '10.11.12.0', 'IPv4 host bits should be zeroed out' ],
+			[ '2001:db8:85a3:8d3:1319:8a2e:370:7348', '2001:db8:85a3::', 'IPv6 address should be truncated to a /48 network' ],
+			[ '2001:db8:85a3::8a2e:370:7334', '2001:db8:85a3::', 'compressed IPv6 address should be truncated to a /48 network' ],
+			[ '2a02:8109:aa00:1234::5', '2a02:8109:aa00::', 'the fourth group should be dropped whole, not truncated to a /56' ],
```
stklcode: One tiny suggestion, I'd replace the only public IP in the test set with a demo or private IP, maybe something from ULA space if we don't want to reuse the `2001:db8::/32` demo net.

```suggestion
			[ 'fd02:8109:aa00:1234::5', 'fd02:8109:aa00::', 'the fourth group should be dropped whole, not truncated to a /56' ],
```

2ndkauboy: Thanks, and done as well!

2ndkauboy:
Squash-merged into `v3` as dd32868. The stacked-PR merge path was blocked by stack #789, which is rooted at the draft PR #506 (`v3` → `master`).

PR #762, opened 2026-07-06
fix: delegate IP anonymization to wp_privacy_anonymize_ip()

2ndkauboy:
## Description

The IPv4 branch of `\AntispamBee\Helpers\IpHelper::anonymize_ip` kept only the first two octets (a `/16`) and the IPv6 branch only the first two groups (a `/32`). A `/16` can span country borders and geolocation databases key country on `/24`, so the previous masking could **degrade the country lookup** in `CountrySpam`, which sends the anonymized IP to iplocate.io (see `src/Rules/CountrySpam.php`).

This delegates to WordPress core's `wp_privacy_anonymize_ip()` when it is available:

- IPv4 → `/24` (first 3 octets), IPv6 → `/48` (first 3 groups) — enough for reliable country-level geolocation while removing the host.
- Core additionally handles ports, brackets and zone identifiers.

A bundled `inet_pton`/`inet_ntop` masking with the **same** `/24` and `/48` masks is kept as a `function_exists()` fallback for environments where the core function does not exist.

## Comparison across versions

`anonymize_ip` has changed masks between versions. This PR **restores the v2 IPv4 behaviour** (`/24`), while making IPv6 more private than v2 but coarse enough to keep country geolocation:

| Version | IPv4 mask | IPv6 mask |
|---|---|---|
| v2 (released) | `/24` (3 octets) | `/16` (1 group) |
| v3 (current) | `/16` (2 octets) | `/32` (2 groups) |
| **This PR** | **`/24` (3 octets)** | **`/48` (3 groups)** |

Worked examples:

| Input | v2 | v3 (current) | This PR |
|---|---|---|---|
| `192.168.1.1` | `192.168.1.0` | `192.168.0.0` | `192.168.1.0` |
| `203.0.113.42` | `203.0.113.0` | `203.0.0.0` | `203.0.113.0` |
| `2001:db8:85a3:8d3:1319:8a2e:370:7348` | `2001::` | `2001:db8::` | `2001:db8:85a3::` |

➡️ **IPv4 anonymization is identical to v2** (`/24`); IPv6 keeps two more groups than v2, restoring reliable country lookups without exposing the host.

## Testing

- Data-provider test covering IPv4/IPv6 (full + compressed) and invalid input, asserting the fallback masking (`wp_privacy_anonymize_ip()` is undefined in the unit env).
- Separate-process test verifying delegation to core when the function exists.

`composer test:unit` — all pass. `phpcs` and `phpstan` clean.

> **Note:** stacked on #761 (the unit-test PR). Base is `test/anonymize-ip-unit-test`; GitHub will auto-retarget to `v3` once #761 merges.

### DIFF
--- a/src/Helpers/IpHelper.php
+++ b/src/Helpers/IpHelper.php
@@ -61,7 +61,17 @@ class IpHelper {
 	}
 
 	/**
-	 * Anonymize the IP addresses
+	 * Anonymize an IP address.
+	 *
+	 * Only the network portion is kept, the host portion is zeroed out. For
+	 * IPv4 the first three octets are retained (a `/24` network), for IPv6 the
+	 * first three groups are retained (a `/48` network). This keeps enough of
+	 * the address for reliable country-level geolocation while removing the host.
+	 *
+	 * WordPress core's {@see wp_privacy_anonymize_ip()} is used when available;
+	 * it applies the same masks and additionally handles ports, brackets and
+	 * zone identifiers. The bundled masking is only a fallback for environments
+	 * where that function does not exist.
 	 *
 	 * @param string $ip Original IP.
 	 *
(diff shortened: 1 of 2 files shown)


src/Helpers/IpHelper.php
```diff
@@ -61,20 +61,58 @@ private static function sanitize_ip( string $raw_ip ): string {
...
 	 *
 	 * @return  string     Anonymous IP.
 	 * @since   2.5.1
 	 */
 	public static function anonymize_ip( string $ip ): string {
-		preg_match( '/\w+([\.:])\w+/', $ip, $matches );
-		$ip_start = $matches[0];
-		if ( '.' === $matches[1] ) {
-			return $ip_start . '.0.0';
+		if ( function_exists( 'wp_privacy_anonymize_ip' ) ) {
```
stklcode, line 82: Function is available since WP 4.9.6 according to the docs:
https://developer.wordpress.org/reference/functions/wp_privacy_anonymize_ip/

We currently require 4.6 in v3, thought is should be 5.x already. Probably not the best reason to raise it, but this may be an option.

stklcode:
Results of WP built-in and fallback are not consistent and the WP implementation truncates to a /64 prefix, not /48 as stated in the comments and PR description.

| Original                             | Custom          | WP                      |
|--------------------------------------|-----------------|-------------------------|
| 192.168.1.1                          | 192.168.1.0     | 192.168.1.0             |
| 10.0.0.1                             | 10.0.0.0        | 10.0.0.0                |
| 2001:db8:85a3:8d3:1319:8a2e:370:7348 | 2001:db8:85a3:: | 2001:db8:85a3:**8d3**:: |
| 2001:db8::1                          | 2001:db8::      | 2001:db8::              |
| not-an-ip                            | not-an-ip       | **0.0.0.0**             |

If we not add additional logic to handle invalid addresses and adjust the fallback to match the WP function, we could also just keep one custom implementation (that hat less features than the WP one, but we don't need port and bracket handling here anyway)

src/Helpers/IpHelper.php
2ndkauboy: I was thinking about a "polyfill" here. Same with another function from 6.2 we would use, if available, and for which we have an alternative.

2ndkauboy:
@stklcode thanks for finding this all out. Then let's not use the Core function for now.

PR #763, opened 2026-07-07
fix: address code review findings from 3.0.0-beta.1

2ndkauboy:
## Summary

This PR fixes the confirmed functional bugs found in a full code review of `3.0.0-beta.1`, one fix per commit:

- **`LinkbackFromMyself` never fired**: `$url[0]` / `$target_post_id[0]` indexed the first *character* of the scalar values from `preprocess_comment`, so the `home_url()` comparison could never match and legitimate self-linkbacks never received their `-100` trust score. (+ new unit tests)
- **Fatal on PHP 7.2**: `Settings` uses `array_key_last()` (PHP 7.3+) while the plugin advertised a PHP 7.2 floor. Since CI only tests PHP 7.4+, the minimum version is now 7.4 across `composer.json`, `phpcs.xml`, the runtime guard in `antispam_bee.php`, `readme.txt` and `AGENTS.md`.
- **`Honeypot::inject()` fails safe now**: it returned an empty string when the comment field markup had no element with the configured `field_id` — themes without `id="comment"` lost their comment textarea entirely. It also fataled on fields without a `name` attribute and emitted libxml warnings on HTML5 markup. All three cases now return the original markup unchanged / stay silent. `get_secret_id_for_post()`/`get_secret_name_for_post()` share one implementation now (output unchanged). (+ new unit tests)
- **`IpHelper::anonymize_ip()`**: empty or unmatchable input (e.g. `::1`) caused undefined-index warnings and a garbage return value; it returns an empty string now. (+ new unit tests)
- **`Select` admin field**: output was completely unescaped and the stored value was never marked as `selected`, so the setting did not survive a settings-page round-trip; multi-selects also submitted only one value (missing `[]` in the name). Latent — no bundled component uses `select` yet — but a landmine for extenders. The dead `Text::get_placeholder()` override (always-unset property shadowing the working parent method) is removed as well.
- **`DataHelper::get_values_where_key_contains()`**: truthy `strpos()` check missed matches at position 0.
- **Debug log disclosure**: the opt-in debug log was written to `wp-content/asb-debug.{date}.log` — a predictable, usually web-accessible path — and contains comment content. The file name now includes a salted hash suffix.

## Test steps

1. `composer test:unit` — 29 tests pass (10 of them new).
2. Honeypot regression: activate a theme whose comment form textarea has no `id="comment"` → the comment field must still render (previously it disappeared).
3. Self-linkback: enable pingbacks, publish post B linking post A, verify the pingback on A is not treated as spam by other rules more aggressively than before (rule now contributes its `-100` trust score).
4. Debug mode: define `ANTISPAM_BEE_DEBUG_MODE_ENABLED`, submit a comment, and verify the log file in `wp-content/` now carries a hash suffix.
5. On a PHP < 7.4 host the plugin deactivates gracefully with the admin notice instead of fataling in `Settings`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -22,9 +22,9 @@ treating it as project code produces false findings.
 
 Fix code style violations automatically with `composer csfix`.
 
-Target PHP 7.2+. Avoid syntax introduced in PHP 7.4 or later: typed properties, arrow functions (
-`fn =>`), `match` expressions, null-coalescing assignment (`??=`). The `phpcs.xml` ruleset enforces
-this via PHPCompatibilityWP.
+Target PHP 7.4+. Avoid syntax introduced in PHP 8.0 or later: `match` expressions, constructor
+property promotion, named arguments, nullsafe operator (`?->`), union types. The `phpcs.xml` ruleset
+enforces this via PHPCompatibilityWP.
 
 ## WordPress & security conventions
 
--- a/antispam_bee.php
+++ b/antispam_bee.php
@@ -35,7 +35,7 @@ pre_init();
  */
 function pre_init(): void {
 	// Check if the min. required PHP version is available and if not, show an admin notice.
-	if ( version_compare( PHP_VERSION, '7.2', '<' ) ) {
+	if ( version_compare( PHP_VERSION, '7.4', '<' ) ) {
 		add_action( 'admin_notices', __NAMESPACE__ . '\min_php_version_error' );
 
 		// Stop the further processing of the plugin.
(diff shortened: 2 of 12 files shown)


2ndkauboy:
Rebased onto `v3` after the code-cleanup stack (#765/#766/#767/#769) landed. Two of the original commits were dropped as **already resolved by #767's normalized rule payload**:

- *remove string char indexing in `LinkbackFromMyself`* — `verify()` now reads scalar `url`/`post_id` from the normalized payload; the regression test lives in #773.
- *strict `strpos()` in `DataHelper::get_values_where_key_contains()`* — that method was removed entirely by the normalized-payload refactor.

The remaining 5 fixes were reconciled against the refactored code (Honeypot fail-safe merged onto the phpstan-safe structure; `Text::get_placeholder()` shadowing bug removed even though code-cleanup had relocated it). PHPStan level 8, unit tests, and phpcs all pass.

PR #766, opened 2026-07-10
fix: resolve all PHPStan level 8 issues

2ndkauboy:
## Summary

Resolves all PHPStan level 8 findings across `src/` and `antispam_bee.php` while keeping PHPCS green. The bulk is adding precise iterable value types to docblocks; the rest fixes the type/logic problems that level 8 surfaced.

## Key changes

- **Type hints:** `array<string, mixed>` for reaction/item data, `string[]` for reason/type lists, `array<int, array<string, mixed>>` for option configs, and `array<class-string<Controllable>>` for component lists. `@phpstan-param` annotations are used instead of `// phpcs:ignore`.
- **Component pipeline:** threaded `class-string<Controllable>` through `ComponentsHelper::filter` and the `get_controllables()` methods; `get()`/`get_spam_reason_rules()` keep the wider `array<class-string>`. Corrected `SettingsPage::$rules`/`$post_processors`, which hold class-strings, not instances.
- **Dynamic constants:** `ANTISPAM_BEE_DEBUG_MODE_ENABLED` and `ANTISPAM_BEE_LOG_FILE` are declared under `dynamicConstantNames` so their user-defined values are not narrowed to the bootstrap literals.
- **Latent bugs fixed:** DOM null-safety and argument types in `Honeypot`; unchecked `preg_match` in `IpHelper::anonymize_ip`; null spam-reason mapping key in `PluginUpdate::convert_multiselect_values`; null `$reaction_type` guard in `ComponentsHelper`; `comment_ID` cast in `Reaction`; `esc_url` protocols and `preg_split` handling in `LangSpam`/`CountrySpam`; offset/`iconv` safety in `RegexpSpam`; the dead `Text::get_placeholder()` override.
- **Regression fix:** `DashboardWidgets::get_spam_count()` reads `spam_count` from the options root (`get_option( 'spam_count', '' )`), consistent with the `UpdateSpamCount` writer. A prior edit had dropped the argument, defaulting `$reaction_type` to `general` and always reading `0`.

## Test steps

- `composer run lint-php` (PHPCS) — green
- `vendor/bin/phpstan analyse` (level 8) — `[OK] No errors`
- `npm run test:e2e` — 34 passed, 2 skipped (v3-removed option), 0 failed. Notably covers the dashboard spam counter (`more.spec.ts`).

### DIFF
--- a/src/Admin/DashboardWidgets.php
+++ b/src/Admin/DashboardWidgets.php
@@ -29,9 +29,9 @@ class DashboardWidgets {
 	/**
 	 * Display the spam counts on the dashboard.
 	 *
-	 * @param array $items Initial array with dashboard items.
+	 * @param string[] $items Initial array with dashboard items.
 	 *
-	 * @return  array Merged array with dashboard items.
+	 * @return  string[] Merged array with dashboard items.
 	 */
 	public static function add_dashboard_count( array $items = [] ): array {
 		if ( ! current_user_can( 'manage_options' ) || ! Statistics::is_active() ) {
@@ -83,6 +83,13 @@ class DashboardWidgets {
 	 * Return the number of spam comments.
 	 */
 	private static function get_spam_count(): int {
-		return intval( Settings::get_option( 'spam_count', 0 ) );
+		return intval( Settings::get_option( 'spam_count', '' ) );
+	}
+
+	/**
+	 * Output the number of spam comments.
+	 */
+	public static function the_spam_count(): void {
+		echo esc_html( (string) self::get_spam_count() );
 	}
 }
--- a/src/Admin/DashboardWidgets.php
+++ b/src/Admin/DashboardWidgets.php
@@ -21,7 +21,6 @@ class DashboardWidgets {
 	 */
 	public static function init(): void {
 		if ( DashboardHelper::is_dashboard_page() ) {
-			add_action( 'antispam_bee_count', [ __CLASS__, 'the_spam_count' ] );
 			add_filter( 'dashboard_glance_items', [ __CLASS__, 'add_dashboard_count' ] );
 		}
 	}
@@ -84,12 +83,5 @@ class DashboardWidgets {
 	 */
 	private static function get_spam_count(): int {
 		return intval( Settings::get_option( 'spam_count', '' ) );
-	}
-
-	/**
-	 * Output the number of spam comments.
-	 */
-	public static function the_spam_count(): void {
-		echo esc_html( (string) self::get_spam_count() );
 	}
 }
--- a/src/Rules/RegexpSpam.php
+++ b/src/Rules/RegexpSpam.php
@@ -29,7 +29,9 @@ class RegexpSpam extends ControllableBase implements SpamReason {
 	 *
 	 * Content fields using pre-defined and custom regular expressions.
 	 *
-	 * @param array{
+	 * @param array<string, mixed> $item Item to verify.
+	 *
+	 * @phpstan-param array{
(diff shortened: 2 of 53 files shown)


2ndkauboy:
@threadi if you want to read some PHPStan fixes, I would really appreciate your help. :pray:

src/Admin/DashboardWidgets.php
```diff
@@ -88,6 +88,16 @@ public static function add_dashboard_count( array $items = array() ): array {
 	 * @since  2.4
 	 */
 	private static function get_spam_count(): int {
-		return intval( Settings::get_option( 'spam_count', 0 ) );
+		return intval( Settings::get_option( 'spam_count', '' ) );
```
stklcode, line 85: I know `intval( 0 ) === intval( '' ) === intval( '0' )`, but I'd prefer the explicit zero here to make the intent clear.

threadi:
> @threadi if you want to read some PHPStan fixes, I would really appreciate your help. 🙏

That looks really nice :) The test is included in the test workflow, too - very nice :)

src/Rules/RegexpSpam.php
```diff
@@ -29,7 +29,17 @@ class RegexpSpam extends ControllableBase implements SpamReason {
...
+	 *
+	 * @phpstan-param array{
+	 *     reaction_type: string,
+	 *     comment_author_IP?: string,
+	 *     comment_author_url?: string,
+	 *     comment_content?: string,
+	 *     comment_author_email?: string,
+	 *     comment_author?: string,
+	 *     comment_agent?: string,
+	 * } $item
```
stklcode, line 42: I personally prefer the WordPress-style DocBlock syntax[^1] like this in a WP project
```php
	 * @param array $item {
	 *     Item to verify.
	 *
	 *     @type string      $reaction_type
	 *     @type string|null $comment_author_IP
	 *     @type string|null $comment_author_url
	 *     @type string|null $comment_content
	 *     @type string|null $comment_author_email
	 *     @type string|null $comment_author
	 *     @type string|null $comment_agent
	 *  }
```

But I'm not aware of any setting to make PHPStan understand these kind of complex type hints and both are little too much. So probably keep the proposed hints.

[^1]: https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#1-1-parameters-that-are-arrays

src/Admin/DashboardWidgets.php
threadi: The second parameter of `get_option` must be a string, not an int. The alternative would be to change its parameter to `int|string` just for this one case, but that would require changing it back to `string` within the functions or performing additional checks for int values. Therefore, this code here is the best solution.

stklcode: I know why this was changed, yet `'0'` is a string that makes clear that "zero" is the intended fallback.

2ndkauboy: Looking at the function name, one could think that the second parameter is the fallback, but it is not. :see_no_evil: 

It is the `$reaction_type`. This code have been changed in a recent other branch.

src/Rules/RegexpSpam.php
2ndkauboy: I can try. Maybe PhpStorm does not mess up this PHPDoc comment. :crossed_fingers:

src/Admin/DashboardWidgets.php
stklcode: Oh, you are right. OK in this case the 0 was always kind of wrong here and empty string should be fine.

Issue #780, opened 2026-07-26
E2E language API mock does not preprocess the text like the real service


2ndkauboy:
The E2E mock in `tests/e2e/language-api/server.js` passes the comment to `franc` unchanged, while the real service at `https://api.pluginkollektiv.org/language/v1/` preprocesses the text first — at least it strips URLs.

Measured with the same input, `感謝分享，這個網站真的很棒 http://example.com/spam`:

* production API → `cmn`
* mock and plain `franc` → `fra`

That makes the mock a poor stand-in for exactly the mixed-content cases the language rule has to get right, and it forces the E2E specs to avoid URLs in their comment texts to stay meaningful (see the note added in #776).

Two things worth doing:

1. Document what the service actually does to the text before detection.
2. Reproduce that preprocessing in the mock, so the E2E tests exercise the same behaviour as production.

Issue #779, opened 2026-07-26
Detect languages of scripts without word delimiters locally


2ndkauboy:
Follow-up to #776.

In #776 it was suggested that even a very short comment such as `调用` should be recognized. That is not reachable through the detection service: `franc` returns `und` for any input shorter than ten characters, before it looks at the script at all (`franc/index.js`, `value.length < minLength`, `minLength = 10`). So the character threshold of the new guard cannot be lowered any further without the service simply answering "undetermined".

It *is* reachable locally, though. `franc` itself identifies these scripts by the script alone, not by comparing character sequences, and `AntispamBee\Helpers\TextHelper` already knows how much of a text belongs to such a script. The rule could map the dominant script itself and skip the request entirely:

* Han → `zh`
* Hiragana or Katakana present → `ja`
* Hangul → `ko`
* Thai → `th`, Lao → `lo`, Khmer → `km`, Myanmar → `my`, Tibetan → `bo`

Benefits:

* Works down to two characters.
* No request to a third-party service for those comments, which is a privacy improvement — the rule description links to a privacy notice for exactly that reason.
* No `und` case to handle, and fully testable offline.

Caveats worth discussing:

* It replicates the Han/Kana ambiguity `franc` has for Japanese: a Japanese text written only in Kanji is indistinguishable from Chinese.
* It is a behaviour change for every comment in such a script, not just the short ones, if the local mapping takes precedence over the service.

PR #781, opened 2026-07-26
fix: detect languages in scripts without word delimiters (#776)

### LINKED ISSUE #776: Language spam word count issue
**Describe the bug**
I’ve got a spam comment in v3 that was not recognized as spam despite only allowing German and English comments, which made me curios.

The comment was:

> 致力于为开发者提供快速、便捷的企业级AI接口调用方案，打造稳定且易于使用的 API 接口平台，一站式集成几乎所有 AI 大模型。

The problem: Since the language detection may not be that accurate with just a few words, we decided to go with a minimum of 10 words before the language API is even used. While that may work for European languages, it totally fails in such cases, because Kanji usually use way less spaces, which is used to determine words in our case.

**To Reproduce**
1. Allow only German and English comments
2. Write the comment above as guest
3. See that it didn’t trigger the language API check

**Expected behavior**
Instead of only relying on words, add a length as well, e.g. 30 – 50 characters. Maybe we can also lower the threshold with https://github.com/pluginkollektiv/antispam-bee/issues/519 even further.

2ndkauboy:
Fixes #776.

## Problem

The language rule only sends a comment to the detection service if it contains enough text, because the service is unreliable for short texts. Whether that amount was measured in words or in characters was decided by `wp_get_word_count_type()` and `get_option( 'blog_charset' )` — so by the locale of the **site**, not by the script of the **comment**.

On a site that allows only German and English, the Chinese comment from the report (66 characters) therefore counted as a single word, never reached the service, and was never recognized as spam.

## Solution

The guard now looks at the comment itself. Letters of scripts that do not delimit their words with spaces (Han, Hiragana, Katakana, Hangul, Thai, Lao, Khmer, Myanmar, Tibetan) are counted as characters, everything else keeps being counted as words:

* **at least 10 code points** — this mirrors the minimum length of the service itself. `franc` returns `und` for anything shorter, before it even looks at the script (`value.length < minLength`, `minLength = 10`), so a lower threshold would only ever produce "undetermined".
* **at least half of all letters** have to belong to such a script — otherwise the service detects the language of the dominant, space delimited part of the text, for which those few characters are just noise.

`wp_get_word_count_type()`, the `_x( 'words', … )` fallback for WordPress < 6.2 and `get_option( 'blog_charset' )` are gone. The last one was also a real `preg_match(): Passing null` deprecation on PHP 8.1 when the option is missing.

Measured against both the production API and the exact `franc` version the E2E mock uses:

| Comment | Before | After |
| --- | --- | --- |
| The Chinese comment from #776 (64 characters) | not checked | sent, flagged |
| `中文垃圾` (4 characters) | not checked | not checked |
| `Great article thanks for sharing 中文垃圾评论` | not checked | not checked |
| `A small text passes the test. Lets check this.` (9 words) | not checked | not checked |

The third row is the reason for the dominance rule: the service answers `dan` (Danish) for that text and `bem` for a comparable mix, so sending it would flag English comments as spam on an English-only site.

## Two defects fixed along the way

Both of these are latent today, because the ten word threshold hides them — and both would have become live false-positive sources the moment the threshold got lowered:

* **`und` was treated as a language.** The service returns the ISO 639-3 code `und` if it could not determine the language, and really does so in production. `LangHelper::map()` passes unknown codes through, so `und` was compared to the allowed languages and **every comment the service could not identify was marked as spam**. It now returns the neutral result, like every other case in which the rule cannot decide.
* **`cmn` was not mapped to `zh`.** The service reports Mandarin as `cmn`, and only `zho` was mapped, so allowing Chinese via `antispam_bee_get_allowed_translate_languages` had no effect. The other members of the Chinese macrolanguage without an ISO 639-1 code of their own (`yue`, `wuu`, `nan`, `hak`) are now mapped as well.

## Structure

The mechanical counting lives in a new `AntispamBee\Helpers\TextHelper`, which makes no WordPress calls at all and is therefore testable without any stubs. The thresholds and the policy stay in the rule and can be adjusted with two new filters, `antispam_bee_lang_min_words` and `antispam_bee_lang_min_characters`.

One detail worth pointing out: PCRE counts the ideographic full stop `。` and the ideographic comma `、` as `\p{Han}`, while they are not letters. The pattern for scripts without word delimiters therefore carries a `(?=[\p{L}\p{M}])` lookahead, so its result stays comparable to the number of letters of the whole text and cannot be inflated with punctuation.

URLs are deliberately **not** stripped before the check, even though the production service strips them. The E2E mock does not, so the same input yields `cmn` in production and `fra` locally. Stripping client side would make the guard more permissive than the detector actually is, so comments mixing a script without word delimiters with URLs are skipped — the conservative direction for a spam filter. See #780.

## Tests

* `tests/Unit/Helpers/TextHelperTest.php` — normalization (including non-breaking spaces and invalid UTF-8) and all counters per script.
* `tests/Unit/Rules/LangSpamTest.php` — the rule had no unit test at all so far. Covers the case from #776, the short and the latin-dominant cases (asserting that **no** request is made), the `und` regression, the `cmn` mapping, both new filters, and that `antispam_bee_detected_lang` still short-circuits the whole check.
* `tests/e2e/specs/filter.spec.ts` — three cases added, mirroring the unit tests end to end. The existing "skips short comments" case is unaffected, its text contains no such characters.

`composer test:unit`, `composer lint-php` and `composer phpstan` are clean. The E2E suite was not run locally (`wp-env` would have contended for ports with another running project), only the spec collection was verified.

## Follow-ups

* #779 — recognizing very short comments such as `调用` locally, which is impossible through the service.
* #780 — the E2E mock does not preprocess the text like the real service.

### DIFF
--- a/src/Rules/LangSpam.php
+++ b/src/Rules/LangSpam.php
@@ -10,6 +10,7 @@ namespace AntispamBee\Rules;
 use AntispamBee\Helpers\LangHelper;
 use AntispamBee\Helpers\Sanitize;
 use AntispamBee\Helpers\Settings;
+use AntispamBee\Helpers\TextHelper;
 use AntispamBee\Interfaces\SpamReason;
 
 /**
@@ -61,32 +62,7 @@ class LangSpam extends ControllableBase implements SpamReason {
 			return (int) ! in_array( $detected_language, $allowed_languages, true );
 		}
 
-		$text = trim( preg_replace( "/[\n\r\t ]+/", ' ', $comment_text ) ?? '', ' ' );
-
-		if ( function_exists( 'wp_get_word_count_type' ) ) {
-			$word_count_type = wp_get_word_count_type();
-		} else {
-			/*
-			 * translators: If your word count is based on single characters (e.g. East Asian characters),
-			 * enter 'characters_excluding_spaces' or 'characters_including_spaces'. Otherwise, enter 'words'.
-			 * Do not translate into your own language.
-			 */
-			// phpcs:ignore WordPress.WP.I18n.MissingArgDomain
-			$word_count_type = _x( 'words', 'Word count type. Do not translate!' );
-		}
-
-		if ( strpos( $word_count_type, 'characters' ) === 0 && preg_match(
-			'/^utf\-?8$/i',
-			get_option( 'blog_charset' )
-		) ) {
-			preg_match_all( '/./u', $text, $words_array );
-			$word_count = count( $words_array[0] );
-		} else {
-			$words_array = preg_split( "/[\n\r\t ]+/", $text, -1, PREG_SPLIT_NO_EMPTY ) ?: [];
-			$word_count  = count( $words_array );
-		}
-
-		if ( $word_count < 10 ) {
+		if ( ! self::has_enough_text_for_detection( $comment_text ) ) {
 			return 0;
 		}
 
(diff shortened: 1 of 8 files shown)


MatzeKitt (approved):

src/Rules/LangSpam.php
```diff
@@ -120,13 +96,77 @@ public static function verify( array $item ): int {
...
+	 * Languages that delimit their words with spaces are measured in words. Scripts
+	 * that do not use spaces to delimit words, for example the Chinese, Japanese,
+	 * Korean or Thai script, are measured in characters instead, because such a text
+	 * would otherwise be counted as a single, far too short word.
+	 *
+	 * @param string $text The text to check.
+	 *
+	 * @return bool Whether the text contains enough content.
+	 */
+	private static function has_enough_text_for_detection( string $text ): bool {
```
2ndkauboy, line 129: **question**: Maybe add some tests for this function as well?

2ndkauboy: Test was added.

2ndkauboy:
## The single flip is a true positive

The comparison above reports one spam/ham flip, `f78a4bedb3f97a03` — ham in `v3`, spam (`asb-lang-spam`) with this branch. I identified that comment in the corpus to check whether it is a genuine catch or a false positive.

**It is a comment a human had already marked as spam by hand.** Its historical label in the corpus is `manually`, i.e. the site owner removed it manually because none of the automatic rules caught it.

| | verdict |
|---|---|
| site owner, by hand | **spam** |
| `v3` | not spam — held for moderation |
| `v3` + this branch | **spam** (`asb-lang-spam`) |

So this change automatically catches a comment that previously required manual moderation. That is exactly the outcome the fix is meant to produce.

### The comment

213 characters of Chinese, attached to a link:

> 这篇文章真的太有用了！我之前也遇到过类似的大型网站性能问题，但一直找不到症结所在。作者详细分析了查询慢的原因，特别是关于评论表和索引的解释，让我一下子就明白了问题出在哪里。尤其是提出的三种解决方案，既有删除无用评论这种直接的方法，也有通过代码过滤评论类型和添加索引这种更彻底的优化思路，非常实用。特别是提到用WP-CLI添加索引的经历，让我意识到有时候解决性能问题需要一些非常规的操作。以后遇到类似情况，我肯定要试试这些方法！

Roughly: *"this article was so useful, I had the same large-site performance problem and could never find the cause, the explanation of the comment table and indexes made it click, especially adding indexes with WP-CLI…"* — content-free praise shaped to the host article, which is the standard form of AI-generated link spam. The author name is Japanese katakana reading **"I'm not human"**, which is a nice touch.

This is precisely the case the rule could not previously see: Chinese has no word delimiters, so before this change the text counted as a single far-too-short "word" and never reached the detection service. Author details and IP omitted here deliberately — the corpus is not public.

### Why the small corpus surfaces it

Both the full-corpus run above and a run against the new ~10k [small corpus](https://github.com/pluginkollektiv/antispam-bee/pull/785) find this same single flip. The small corpus keeps the `manually` stratum **whole** (all 210 comments), just as it keeps all 6,755 ham comments, so a change of this kind cannot be missed by sampling.

### One caveat on reading flips

The corpus's labels are historical, and they are not equally strong evidence in both directions. `manually` is a human decision, so a `manually` → spam flip is a confirmed win. The absence of a label only means nothing flagged the comment at the time — it does not prove the comment was legitimate, so a ham → spam flip needs a look rather than being assumed to be a regression. This flip is the first kind.

Issue #768, opened 2026-07-12
[Integrations] How could 3rd party integrate with us?


Zodiac1978:
This is a ticket for collecting thoughts about how 3rd parties could integrate with us.

# My first thoughts ...

### Requires mapping to an e-mail or username
* Trust approved commenters
* Trust commenters with a Gravatar

### Requires mapping to a textarea
* BBCode links are spam
* Allow comments only in certain language (Privacy!)

### Requires mapping to various fields (body -> textarea, e-mail, IP ...)
* Use regular expressions
* Look in the local spam database

### No mapping needed
* Honeypot
* Consider the comment time
* Block comments from specific countries

### New/better spam checks 
* Timecheck via JS to make it compatible with caching
* Use native Disallowed list from WordPress (Requires mapping to a textarea)
* Rate limiting by IP
* Block bad bots / user agents
* Token (HMAC)

Here is an interesting mapping mechanism for CF7:
https://contactform7.com/spam-filtering-with-akismet/#deploying-akismet-in-a-contact-form

Zodiac1978:
### Which 3rd party would be interesting to have an integration with?

#### Core
* Login
* Registration

#### ContactForm Plugins
* Ninja Forms
* Gravityforms
* WS Form
* WPForms
* Form Block
* Jetpack Forms
* Elementor Forms
* Elementor Atomic Forms
* Everest Forms
* Formidable Forms
* Contact Form 7
* SureForms
* JetFormBuilder
* Forminator Forms
* Fluent Forms
* Bricks Builder Forms
* Breakdance Builder Forms
* Divi Contact Form

#### Transactions / Newsletter
* Brevo
* The Newsletter

#### eCommerce
* WooCommerce (Review/Checkout)

2ndkauboy:
How would you use Antispam Bee for the WordPress login? As some kind of "limit login prevention"? But what would be the decisions to block a request? 🤔

2ndkauboy:
Antispam Bee should already block WooCommerce reviews, since they are a comment type we support.

Zodiac1978:
> How would you use Antispam Bee for the WordPress login? As some kind of "limit login prevention"? But what would be the decisions to block a request? 🤔

Yes, login prevention. Why not? If the login comes from an IP already known in the local spam DB, we could block it. We could add a honeypot, to prevent the login. We could block logins from specific countries. We could block bad bots/user agents. We could use a (one-time) token system to prevent direct form injection.

[...]

PR #795, opened 2026-08-07
feat: add a public spam check API for third-party integrations

### LINKED ISSUE #507: Add docs how to extend ASB 3
- How to add custom rules

2ndkauboy:
Closes #507.

Adds a documented, stable entry point for plugins that want their own content checked, so `AntispamBee\Handlers\Rules` can stay internal.

The trigger was the first real-world integration, [epiphyt/form-block on its `antispam-bee` branch](https://github.com/epiphyt/form-block/tree/antispam-bee). Reviewing it surfaced three things worth settling *before* writing documentation, because the docs freeze whatever we describe.

## 1. `new Rules( $reaction_type )` was the only way in

`Rules` is an internal handler that `Reaction::process()` instantiates. Documenting it would have made its constructor permanent third-party API.

`AntispamBee\Api\SpamCheck` is now the entry point, and `Handlers\Rules` and `Handlers\PostProcessors` are marked `@internal`:

```php
$result = \AntispamBee\Api\SpamCheck::check(
    [ 'author' => $name, 'body' => $message, 'email' => $email ],
    'my_plugin_form'
);

if ( $result->is_spam() ) {
    $reasons = $result->get_reasons();
}
```

`CheckResult` carries the verdict, the reason slugs and the payload. It is an object rather than a boolean mainly because of `was_evaluated()`: a result can report "no spam" simply because no rule was active, and that is the most common reason an integration appears to do nothing. `has_active_rules()` lets an integration warn an administrator before any submission arrives, and `get_reason_texts()` resolves stored slugs so integrators need a single import.

## 2. The payload contract was easy to get wrong

form-block hand-built the item and left `host` empty. `RegexpSpam` reads `host` separately from `rawurl`, so a good number of our own patterns could never match. Nobody should have to know that `host` is `wp_parse_url( $url, PHP_URL_HOST )`.

`SpamCheck::check()` normalizes: it derives `host` from `url` via `DataHelper::parse_url()` and defaults `ip` and `useragent` from the current request, so a form plugin passes only what it actually has. The new `antispam_bee_api_payload` filter is the escape hatch.

## 3. Nothing was checked until an administrator ticked a box

`Settings::$defaults` only covered `comment`, `linkback` and `general`, so every `ControllableBase` rule started inactive for a reaction type registered by another plugin — including rules the integration had explicitly opted in. Verified on a local install: form-block's integration worked, but only because its settings tab had been visited and BBCode ticked. `rule_asb_regexp_active` was still unset there, so `RegexpSpam` silently never ran.

`Settings::get_defaults()` applies a new `antispam_bee_default_options` filter, and `get_options()` fills in the defaults of reaction types absent from the stored options.

**Defaults deliberately apply only to reaction types that were never saved.** An unticked checkbox is *removed* from the stored options by `Sanitize::sanitize_controllables()`, so merging defaults underneath the stored state would re-enable a rule an administrator had deliberately disabled — for `comment` and `linkback` too. There is a unit test for exactly this.

`Rules::apply()` now logs when no rule is active, so the "nothing ran" case leaves a trace instead of silently reporting no spam.

## Also: comment-only post-processors

`SaveReason` and `SendEmail` defer their work to `comment_post`. For an item that never becomes a comment the callback not only never runs, it stays attached for the rest of the request — so an unrelated comment inserted later would be stamped with that item's spam reasons, or trigger its notification email. Both now bail unless `comment_post_ID` is set, mirroring what `UpdateSpamLog::process()` already does.

The check is structural on purpose: `comment_type` is legitimately empty for a regular comment, and gating on `reaction_type` would only re-block what an integration explicitly opted in via `antispam_bee_post_processor_supported_types`.

## Documentation

New `docs/integrating-own-content.md` covers the API, the payload attributes, registering a custom reaction type, and registering your own post-processor via `antispam_bee_post_processors` (already supported, previously undocumented). Two traps are called out explicitly:

* Rules register during module initialization, which is skipped for Ajax requests, so a plugin submitting over Ajax must allow it via `antispam_bee_disallow_ajax_calls` **and register that filter early**. REST requests are unaffected, because `wp_doing_ajax()` is false for them.
* Our own post-processors support comments and linkbacks *by design*: they write comment meta, build notifications from a stored comment, and feed a spam counter rendered next to the comment counts in the At a Glance widget. They should not be opted into a custom reaction type. The documentation says so rather than advertising them.

`docs/adding-rules.md` gains the registration step, the reserved `asb-` slug prefix and a note on the payload. `docs/program-flow.md` no longer leaves `Handlers\Rules` looking like the way in.

## Testing

`composer test:unit` — 56 tests pass, 20 of them new. `phpstan` and `phpcs` clean.

Verified end to end against a local install with form-block ported to the new API, submitting real forms over HTTP:

| Check | Result |
|---|---|
| BBCode submission | stored as spam, reason `asb-bbcode` |
| `host` derived from `url` | `spam.example.com` |
| Reaction type with stored options | only the ticked rule active; a new default does **not** re-enable a disabled rule |
| Same reaction type with its stored options removed | all four opted-in rules active from defaults, nothing written to the database |
| Submission only `RegexpSpam` can catch, on defaults alone | stored as spam, reason `asb-regexp` |
| `spam_count` throughout | unchanged at `0` — no post-processor ran, the comment counter is untouched |

## Notes for review

* `SpamCheck::check()` fails open, matching current behaviour for an empty rule set: it catches the `ReflectionException` that `Rules::get()`/`apply()` declare, logs, and returns a not-evaluated result rather than surfacing a Reflection error to integrators.
* `API_VERSION` is there so integrations can adapt to a breaking change; everything outside `AntispamBee\Api` and the documented hooks is stated to be internal.
* The counter stays comment-scoped. A per-reaction-type counter would need new user-facing strings, i18n and a migration for the flat `spam_count`, and nothing requires it yet.

### DIFF
--- a/docs/adding-rules.md
+++ b/docs/adding-rules.md
@@ -1,6 +1,8 @@
 # Adding rules
 
-ASB 3 makes it possible for third-party plugins to add custom rules.
+ASB 3 makes it possible for third-party plugins to add custom rules. If you instead want your own
+content checked by the rules that already exist, see
+[Integrating your own content](integrating-own-content.md).
 
 Each Rule is a class that extends the `ControllableBase` (if your rule has options, like for
 disabling/enabling) or `Base` (if your rule works invisible in the background) class and implements
@@ -16,3 +18,33 @@ honeypot field). Final rules are checked before all other rules, and a positive 
 item as spam immediately, without evaluating the remaining rules. The finals-first ordering can be
 disabled with the `antispam_bee_sort_final_rules_first` filter (return `false`), in which case all
 rules are checked in their registered order.
+
+## Registering a rule
+
+Rules are registered by adding their class name to the `antispam_bee_rules` filter. Extending `Base`
+or `ControllableBase` gives you an `init()` method that does this, so calling it is enough:
+
+```php
+add_action(
+    'plugins_loaded',
+    function (): void {
+        if ( class_exists( '\AntispamBee\Rules\Base' ) ) {
+            \My_Plugin\Rules\My_Rule::init();
+        }
+    }
+);
+```
+
+The `asb-` slug prefix is reserved for the rules that ship with Antispam Bee. Rules from other
+plugins using that prefix are rejected, so pick a prefix of your own.
+
+Which reaction types a rule applies to is declared with the `$supported_types` property, which
+defaults to comments and linkbacks. A rule extending `ControllableBase` is only applied when it is
+active for the reaction type, which means it needs a default or an administrator enabling it — see
+[Integrating your own content](integrating-own-content.md) for how that works.
+
+## Payload
+
+`verify()` receives the normalized payload, not the raw reaction. Its attributes are the same ones
+listed in [Integrating your own content](integrating-own-content.md), so a rule that reads `body` and
+`email` works for comments, linkbacks and any reaction type a third-party plugin registers.
--- a/docs/integrating-own-content.md
+++ b/docs/integrating-own-content.md
@@ -162,13 +162,15 @@ affected by this at all.
 If you enable the [debug mode](https://antispambee.pluginkollektiv.org/documentation/), Antispam Bee
 logs when it finds no active rule for a reaction type, which makes both of these cases visible.
 
-## Reacting to spam the way Antispam Bee does
+## Reacting to spam
 
-`check()` only classifies. It does not count the item, store a reason, notify anybody, or delete
-anything. If you handle storage and notification yourself, that is all you need.
+`check()` only classifies. It does not store anything, notify anybody, or delete anything. Deciding
+what happens to a spam submission is your plugin’s job, and for most integrations doing that directly
+is all you need.
 
-To have Antispam Bee react as it does to its own spam — updating the spam counter and the statistics,
-sending notifications — pass the result to the post-processors:
+If you want that handling to be pluggable — so that site owners or add-ons can change it, and so it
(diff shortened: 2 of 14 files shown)


Zodiac1978:
Related #768

MatzeKitt (commented):
Successfully tested with https://github.com/epiphyt/form-block/pull/83/ 🎉

faisalahammad:
Thanks, the structure with adding-rules and integrating-own-content looks good. A few things I would still want in docs/adding-rules.md before it closes #507:

1. A complete worked example. The page explains the concepts but never shows a full rule class. A minimal `ControllableBase` subclass with `verify()`, `get_name()`, `get_reason_text()` and the registration snippet would take plugin authors from zero to a working rule.
2. `SpamReason` reads as part of the rule contract, but it is optional. `ApprovedEmail` and `ValidGravatar` are `ControllableBase` rules without it. Only the `get_reason_text()` output shows up as a spam reason in the backend.
3. Score semantics: the page says an item is spam if the overall score is positive. The actual check is against the `antispam_bee_spam_threshold` filter (default 0.0), with an early ham return at `antispam_bee_no_spam_threshold`. Both filters deserve a mention.
4. Rule options: `get_options()` format, per-reaction-type activation (`rule_{slug}_active`), `only_print_custom_options()` and `$is_invisible` are not covered. Anyone writing a `ControllableBase` rule will run into these right away.
5. The `antispam_bee_rule_supported_types` filter is undocumented.

Happy to provide a patch for any of these if useful.

Issue #796, opened 2026-08-07
Pass the normalized payload to post-processors


2ndkauboy:
## Problem

Post-processors receive the **raw reaction**, not the normalized payload the rules work on. `Reaction::process()` builds `$payload` via `build_payload()` and hands it to `Rules::apply()`, but `handle_spam()` then calls `PostProcessors::apply( static::$reaction_type, $reaction, … )` with the untouched `comment_*` array.

So a post-processor that wants the client IP has to read `comment_author_IP` (`SendEmail.php:134`, `UpdateSpamLog.php`), while the rules read the normalized `ip` — two names for the same value, and the normalization work (`DataHelper::parse_url()` for `host`, the `IpHelper` lookup, the linkback array-to-scalar flattening in `Linkback::scalar()`) is thrown away before the post-processors run.

It gets worse for third-party integrations: `SpamCheck::post_process()` passes the caller's own array straight through, and `docs/integrating-own-content.md` documents that `process()` receives "the item you passed to `post_process()`". An integrator who fed `SpamCheck::check()` an item with `ip` will reasonably pass that same array on — so a post-processor has no reliable attribute name to read at all. #797 works around this in `UpdateSpamLog` with a `comment_author_IP ?? ip` fallback, which is exactly the wart this issue should remove.

## Proposal

Add the payload to the item instead of replacing it — the raw reaction has to survive:

- `Reaction::handle_spam()` returns `$reaction` into `preprocess_comment`, so WordPress inserts what comes back.
- `SaveReason` and `SendEmail` need the `comment_*` fields to *act* on, not just to read.
- `DeleteForReasons` communicates back through the same array (`asb_marked_as_delete`).

So `PostProcessors::apply()` should inject the payload alongside what it already adds (`asb_reasons`, `reaction_type`), e.g. as `asb_payload`, and the shipped post-processors should read normalized attributes from there. `SpamCheck::post_process()` can inject it just as easily — the `CheckResult` already carries the payload.

That makes every post-processor work off one documented set of attribute names regardless of where the item came from, while remaining backwards compatible for anything reading the raw fields.

## To do

- [ ] Inject the payload in `PostProcessors::apply()`
- [ ] Inject it in `SpamCheck::post_process()` from the `CheckResult`
- [ ] Move `SendEmail` and `UpdateSpamLog` over to the normalized attributes (and drop the `comment_author_IP ?? ip` fallback)
[...]

PR #797, opened 2026-08-07
feat: log the spam reasons in the spam log file

2ndkauboy:
Reported in [this support topic](https://wordpress.org/support/topic/slightly-extended-fail2ban-integration/#post-18987671): the spam log records only *that* something was spam, never *why*, so a Fail2Ban jail cannot ban a honeypot hit longer than a content-based detection.

The log line has been unchanged since 2.5.7 — and, contrary to what the topic reply suggests, `v3` did not make it any more verbose: `UpdateSpamLog::process()` wrote exactly what `_update_spam_log()` writes on `master`.

## What changed

**1. A `key=value` line.** After the discussion below, the format is space-separated pairs rather than the reasons appended to the old sentence:

```
2026-01-15T10:23:45+01:00 ip=192.0.2.42 type=comment post=474 reasons=asb-honeypot
```

Space-separated `key=value` has real precedent in Fail2Ban-land: `filter.d/dovecot.conf` matches `rip=<HOST>` on a very large number of servers. JSONL was considered and rejected for this particular log — Fail2Ban parses with regexes only, a leading `{` forces every operator to configure a custom `datepattern`, and escaped values are a hazard for regex extraction. `antispam_bee_spam_log_entry` makes JSONL about three lines of userland code for anyone who wants it.

Every value is space-free by construction, so nothing needs quoting. Reaction types and rule slugs can come from third-party integrations, so they are reduced to `[a-zA-Z0-9_-]` before being written.

**2. A field that does not apply is written as `-`.** Not omitted — so the field set is identical on every line and a parser never has to cope with a missing key:

```
2026-01-15T10:23:45+01:00 ip=192.0.2.42 type=my_plugin_form post=- reasons=asb-honeypot
```

**3. Reactions other than comments are logged.** Only the IP is essential for a Fail2Ban filter, so an item without `comment_post_ID` is logged with its reaction type and `post=-`. Unlike `SaveReason`/`SendEmail` (see b51a1ed) nothing here is comment-specific — the processor just appends a line. The default `$supported_types` is untouched; this only stops the processor from doing nothing when an integration opts in via `antispam_bee_post_processor_supported_types`.

The IP is read from `comment_author_IP` or, as a fallback, from the normalized `ip` attribute. That fallback is a wart — #796 tracks removing it by handing the normalized payload to post-processors.

**4. The timestamp is ISO 8601 with a UTC offset.** `current_time( 'mysql' )` wrote WordPress-local time without an offset, so on a site whose timezone differs from the server's, Fail2Ban read entries as outside `findtime` and silently never banned. The offset is derived from the difference between local and UTC time at the same instant rather than from the `gmt_offset` option, so it follows daylight saving and cannot disagree with the wall clock it is appended to. The timestamp stays first, because Fail2Ban looks for it near the start of the line.

**5. Two filters: one to extend the line, one to replace it.**

The line is not assembled with a single `sprintf()`. `get_entry()` builds an associative array of fields, filters it, and joins it afterwards — so adding one field does not mean reimplementing the format in userland and keeping it in step with ours by hand:

```php
add_filter(
	'antispam_bee_spam_log_fields',
	function ( array $fields, array $item ): array {
		$fields['score'] = 7.5;

		return $fields;
	},
	10,
	2
);
```

```
2026-01-15T10:23:45+01:00 ip=192.0.2.42 type=comment post=474 reasons=asb-honeypot score=7.5
```

Joining *after* the filter is what makes that safe. Keys are reduced to `[a-zA-Z0-9_-]` and dropped if nothing is left; values have their whitespace replaced and empty ones become `-`; an array value is joined with commas. So an added field cannot split into two fields or corrupt the file.

Values deliberately do **not** go through the slug sanitizer. Only whitespace and control characters break the format, and stripping punctuation would quietly turn a `score=7.5` into `score=75`. Whitespace is also collapsed *before* control characters are stripped — a newline is both, and deleting it outright would run two words together.

The timestamp is not in the array at all: it has no key, and Fail2Ban looks for it near the start of the line.

`ip` *can* be removed. An earlier revision put it back, on the grounds that the documented filter matches on it — but that was the wrong call. Everywhere else this PR guards against *us* breaking someone's setup without telling them; a callback unsetting a field is the site owner instructing us in their own code, and silently ignoring that is worse than the failure it was avoiding. Dropping the address is also a legitimate thing to want: the log sits in `wp-content`, an IP is personal data, and a log kept for statistics rather than for banning has no need of one. The filter documents the consequence instead.

`antispam_bee_spam_log_entry` keeps its meaning — the finished line, for anyone who wants CSV, TSV or JSONL instead of extending ours. Returning an empty string skips the item, e.g. to log honeypot hits only. The entry is filtered without its newline and collapsed to a single line before writing.

## Backward compatibility

**This breaks filters anchored on `marked as spam$`.** That is deliberate, and it is the outcome of the discussion in the comments: 3.0 is where a major release is allowed to break, and a versioned format constant would have meant a value matrix to document and test.

Preserving the old line is **#807's** job, not this PR's. That branch resolves *which* constant supplied the log path, so it can keep `ANTISPAM_BEE_LOG_FILE` writing the byte-identical 2.5.7 line while `ANTISPAM_BEE_SPAM_LOG` gets this format. #807 is stacked directly on this branch, so both land in the same beta and no released version ever breaks a jail.

The changelog carries the warning in English and German.

## Matching the IP

The readme documents `ip=<HOST>` as the field to match on, and promises it keeps its name and shape in future versions of the format. That is the useful guarantee — a version marker like `ASB_SPAM_V1` would not help, because a filter written for V1 meeting a V2 line simply stops matching and Fail2Ban says nothing either way.

```ini
[Definition]
failregex = ^.*?\bip=<HOST>\b
ignoreregex =
```

**The `?` is load-bearing.** Greedy `^.*\bip=<HOST>\b` binds to the *last* `ip=` on the line, so a field added through `antispam_bee_spam_log_fields` whose value contains `ip=` would decide who gets banned — most likely exactly where that value carries attacker-influenced content. The lazy form takes the first, which is ours.

Because the reasons are logged, a jail can also act on one detection only — which is what the original report asked for:

```ini
failregex = ^\s*ip=<HOST> .*\breasons=(?:\S+,)?asb-honeypot(?:,\S+)?(?:\s|$)
```

Worth knowing when writing one: **Fail2Ban strips the timestamp from the line before applying `failregex`**, so an expression anchored on it never matches. The readme says so, and warns that a stricter form spelling out the whole field set is more precise but stops matching if that set changes.

Every expression above was checked with the real `fail2ban-regex` against a log containing multi-reason lines, a non-comment reaction and a line with a hijacking `ip=` in a trailing field: the short filter matches 6/6 with `0 missed`, the honeypot filter matches exactly the three honeypot lines, and the date template is auto-detected **including the zone offset**, so no custom `datepattern` is needed. Shipping a filter as a file in the repo is left to a follow-up — it is a packaging question of its own.

## Deliberately not included

**The score.** `Rules::apply()` does compute one, but it is a local variable that never leaves the method, and today's rule values (`-100`, `-1`, `0`, `1`, `999` with `$weight` always `1`) are a boolean OR with vetoes rather than a calibrated scale. Making the score a real control — configurable weights, a threshold that decides — is planned for 3.1 or later and tracked in #837.

## Test steps

1. Define `ANTISPAM_BEE_LOG_FILE` in `wp-config.php`, pointing at a writable file.
2. Submit a comment that trips the honeypot; the line ends in `reasons=asb-honeypot`.
3. Submit one that trips several rules; the reasons are comma-separated.
4. Submit one through a non-comment integration; it is logged with `type=<reaction type> post=-`.
5. Add a filter on `antispam_bee_spam_log_fields` appending a field; it appears at the end of the line.
6. Add a filter on `antispam_bee_spam_log_entry` returning `''`; nothing is logged.

`composer cs`, `composer phpstan` and `composer test:unit` pass; the post-processor got a unit test suite, it had none before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,9 @@
     * Allows using Antispam Bee rules for other reactions than comments, for example, forms
     * The built-in patterns of the regular expression rule now have stable identifiers as array keys,
       so single patterns can be removed or modified via the `antispam_bee_patterns` filter
-    * The spam log now records the reasons an item was detected and supports reactions other than
-      comments
+    * The spam log now uses a `key=value` format with an ISO 8601 timestamp, records the reasons an
+      item was detected and supports reactions other than comments. Fail2Ban filters written for the
+      previous format need to be updated to match `ip=<HOST>`
     * Fix: The language rule now also checks comments written in a script that does not delimit
       its words with spaces, for example Chinese, Japanese, Korean or Thai
     * Fix: The language rule no longer marks a comment as spam if the language could not be
@@ -38,8 +39,9 @@
     * Die eingebauten Patterns der Regular-Expression-Regel haben nun feste Bezeichner als Array-Keys,
       sodass einzelne Patterns über den Filter `antispam_bee_patterns` entfernt oder angepasst werden
       können
-    * Das Spam-Log enthält jetzt die Gründe der Erkennung und unterstützt auch andere Reaktionen als
-      Kommentare
+    * Das Spam-Log verwendet jetzt ein `key=value`-Format mit einem ISO-8601-Zeitstempel, enthält die
+      Gründe der Erkennung und unterstützt auch andere Reaktionen als Kommentare. Fail2Ban-Filter für
+      das bisherige Format müssen auf `ip=<HOST>` umgestellt werden
     * Fix: Die Sprach-Regel prüft nun auch Kommentare in einer Schrift, die ihre Wörter nicht
       durch Leerzeichen trennt, beispielsweise Chinesisch, Japanisch, Koreanisch oder Thai
     * Fix: Die Sprach-Regel markiert einen Kommentar nicht mehr als Spam, wenn die Sprache
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
     * Allows using Antispam Bee rules for other reactions than comments, for example, forms
     * The built-in patterns of the regular expression rule now have stable identifiers as array keys,
       so single patterns can be removed or modified via the `antispam_bee_patterns` filter
+    * The spam log now records the reasons an item was detected and supports reactions other than
+      comments
     * Fix: The language rule now also checks comments written in a script that does not delimit
       its words with spaces, for example Chinese, Japanese, Korean or Thai
     * Fix: The language rule no longer marks a comment as spam if the language could not be
@@ -36,6 +38,8 @@
     * Die eingebauten Patterns der Regular-Expression-Regel haben nun feste Bezeichner als Array-Keys,
       sodass einzelne Patterns über den Filter `antispam_bee_patterns` entfernt oder angepasst werden
       können
+    * Das Spam-Log enthält jetzt die Gründe der Erkennung und unterstützt auch andere Reaktionen als
+      Kommentare
     * Fix: Die Sprach-Regel prüft nun auch Kommentare in einer Schrift, die ihre Wörter nicht
       durch Leerzeichen trennt, beispielsweise Chinesisch, Japanisch, Koreanisch oder Thai
     * Fix: Die Sprach-Regel markiert einen Kommentar nicht mehr als Spam, wenn die Sprache
--- a/readme.txt
+++ b/readme.txt
@@ -122,7 +122,17 @@ Yes. Define the constant `ANTISPAM_BEE_LOG_FILE` in your `wp-config.php` with th
 
 The fields are space-separated `key=value` pairs, and every value is free of spaces, so the line can be parsed without quoting. A field that does not apply to a reaction is written as `-`; a form submission, for example, has no post to refer to. `reasons` lists the slugs of the rules that caught the item, so a Fail2Ban jail can ban honeypot hits longer than other detections.
 
-Match the IP with `ip=<HOST>` rather than by position: that field is the one Fail2Ban needs, and it will keep its name and shape in future versions of the format.
+Match the IP with `ip=<HOST>` rather than by position: that field is the one Fail2Ban needs, and it will keep its name and shape in future versions of the format. This filter bans every detected item and keeps working if the rest of the line changes:
+
+`failregex = ^.*?\bip=<HOST>\b`
+
+The `?` matters. Without it the expression is greedy and binds to the *last* `ip=` on the line, so a field added by a plugin whose value happens to contain `ip=` would decide who gets banned.
+
+Because the reasons are logged, a jail can also act on one detection only — banning honeypot hits, which no human ever triggers, for far longer than a content-based match:
+
+`failregex = ^\s*ip=<HOST> .*\breasons=(?:\S+,)?asb-honeypot(?:,\S+)?(?:\s|$)`
+
+Two things to know when writing your own. Fail2Ban removes the timestamp from the line before applying `failregex`, so do not try to match it — anchor on `ip=` instead. And a stricter expression that spells out the whole line, such as `^\s*ip=<HOST> type=\S+ post=\S+ reasons=\S+`, is more precise but will stop matching if the field set ever changes; the short form above is the one to prefer unless you need the precision.
 
 To add a field of your own, use the `antispam_bee_spam_log_fields` filter: it receives the fields as an associative array, and anything you append becomes another `key=value` pair at the end of the line. Keys and values are cleaned up afterwards, so an added field cannot break the format.
 
--- a/readme.txt
+++ b/readme.txt
@@ -115,6 +115,13 @@ No, Antispam Bee is free forever, for both private and commercial projects. You 
 
 Complete documentation is available on [pluginkollektiv.org](https://antispambee.pluginkollektiv.org/documentation/).
 
+### Can I log detected spam for Fail2Ban? ###
+Yes. Define the constant `ANTISPAM_BEE_LOG_FILE` in your `wp-config.php` with the path to a writable file. Antispam Bee then appends one line per detected spam item:
+
+> 2026-01-15 10:23:45 comment for post=474 from host=192.0.2.42 marked as spam (asb-honeypot)
+
+The reasons in brackets are the slugs of the rules that caught the item, so a Fail2Ban jail can, for example, ban honeypot hits longer than other detections. Items that are not comments are logged with their reaction type instead of the post reference. Use the `antispam_bee_spam_log_entry` filter to write your own format, or return an empty string from it to skip an item.
+
 ### How can I report security bugs? ###
 You can report security bugs through the Patchstack Vulnerability Disclosure Program. The Patchstack team helps validate, triage and handle any security vulnerabilities. [Report a security vulnerability.](https://patchstack.com/database/vdp/445425e4-f5dd-4404-80a7-690999f5bcb3)
 
@@ -124,6 +131,7 @@ You can report security bugs through the Patchstack Vulnerability Disclosure Pro
     * Complete code rewrite and backend UI overhaul
     * Allows extending Antispam Bee with your own rules
     * Allows using Antispam Bee rules for other reactions than comments, for example, forms
+    * The spam log now records the reasons an item was detected and supports reactions other than comments
     * Fix: The language rule now also checks comments written in a script that does not delimit its words with spaces, for example Chinese, Japanese, Korean or Thai
     * Fix: The language rule no longer marks a comment as spam if the language could not be determined at all
     * Fix: IP addresses are now anonymized by masking the host portion of the address: a /24 for IPv4, the same network WordPress itself keeps, as in Antispam Bee 2.x, and a /48 for IPv6 instead of only its first two groups
--- a/readme.txt
+++ b/readme.txt
@@ -124,7 +124,9 @@ The fields are space-separated `key=value` pairs, and every value is free of spa
 
 Match the IP with `ip=<HOST>` rather than by position: that field is the one Fail2Ban needs, and it will keep its name and shape in future versions of the format.
 
-Use the `antispam_bee_spam_log_entry` filter to write your own format, or return an empty string from it to skip an item.
+To add a field of your own, use the `antispam_bee_spam_log_fields` filter: it receives the fields as an associative array, and anything you append becomes another `key=value` pair at the end of the line. Keys and values are cleaned up afterwards, so an added field cannot break the format.
+
+To replace the line wholesale — with CSV, JSON or anything else — use the `antispam_bee_spam_log_entry` filter, or return an empty string from it to skip an item.
 
 ### How can I report security bugs? ###
 You can report security bugs through the Patchstack Vulnerability Disclosure Program. The Patchstack team helps validate, triage and handle any security vulnerabilities. [Report a security vulnerability.](https://patchstack.com/database/vdp/445425e4-f5dd-4404-80a7-690999f5bcb3)
--- a/readme.txt
+++ b/readme.txt
@@ -118,9 +118,13 @@ Complete documentation is available on [pluginkollektiv.org](https://antispambee
 ### Can I log detected spam for Fail2Ban? ###
 Yes. Define the constant `ANTISPAM_BEE_LOG_FILE` in your `wp-config.php` with the path to a writable file. Antispam Bee then appends one line per detected spam item:
 
-> 2026-01-15 10:23:45 comment for post=474 from host=192.0.2.42 marked as spam (asb-honeypot)
+> 2026-01-15T10:23:45+01:00 ip=192.0.2.42 type=comment post=474 reasons=asb-honeypot
 
-The reasons in brackets are the slugs of the rules that caught the item, so a Fail2Ban jail can, for example, ban honeypot hits longer than other detections. Items that are not comments are logged with their reaction type instead of the post reference. Use the `antispam_bee_spam_log_entry` filter to write your own format, or return an empty string from it to skip an item.
+The fields are space-separated `key=value` pairs, and every value is free of spaces, so the line can be parsed without quoting. A field that does not apply to a reaction is written as `-`; a form submission, for example, has no post to refer to. `reasons` lists the slugs of the rules that caught the item, so a Fail2Ban jail can ban honeypot hits longer than other detections.
+
+Match the IP with `ip=<HOST>` rather than by position: that field is the one Fail2Ban needs, and it will keep its name and shape in future versions of the format.
+
+Use the `antispam_bee_spam_log_entry` filter to write your own format, or return an empty string from it to skip an item.
 
 ### How can I report security bugs? ###
 You can report security bugs through the Patchstack Vulnerability Disclosure Program. The Patchstack team helps validate, triage and handle any security vulnerabilities. [Report a security vulnerability.](https://patchstack.com/database/vdp/445425e4-f5dd-4404-80a7-690999f5bcb3)
@@ -131,7 +135,7 @@ You can report security bugs through the Patchstack Vulnerability Disclosure Pro
     * Complete code rewrite and backend UI overhaul
     * Allows extending Antispam Bee with your own rules
     * Allows using Antispam Bee rules for other reactions than comments, for example, forms
-    * The spam log now records the reasons an item was detected and supports reactions other than comments
+    * The spam log now uses a `key=value` format with an ISO 8601 timestamp, records the reasons an item was detected and supports reactions other than comments. Fail2Ban filters written for the previous format need to be updated to match `ip=<HOST>`
     * Fix: The language rule now also checks comments written in a script that does not delimit its words with spaces, for example Chinese, Japanese, Korean or Thai
     * Fix: The language rule no longer marks a comment as spam if the language could not be determined at all
     * Fix: IP addresses are now anonymized by masking the host portion of the address: a /24 for IPv4, the same network WordPress itself keeps, as in Antispam Bee 2.x, and a /48 for IPv6 instead of only its first two groups
--- a/src/PostProcessors/UpdateSpamLog.php
+++ b/src/PostProcessors/UpdateSpamLog.php
@@ -90,11 +90,11 @@ class UpdateSpamLog extends Base {
 	 * Space-separated `key=value` pairs, the shape `filter.d/dovecot.conf` already
 	 * matches on a great many servers, so it is not exotic in Fail2Ban terms. Every
 	 * value is space-free by construction, which is what makes the line parseable
-	 * without quoting. A field that does not apply to a reaction is written as `-`
+	 * without quoting. A field that has no value for a reaction is written as `-`
 	 * rather than omitted, so the field set is the same on every line.
 	 *
-	 * The timestamp comes first because Fail2Ban looks for it near the start of the
-	 * line to apply `findtime`.
+	 * The timestamp is not one of the fields: it has no key, and Fail2Ban looks for
+	 * it near the start of the line to apply `findtime`, so it must not be movable.
 	 *
 	 * @param array<string, mixed> $item Item that was marked as spam.
 	 * @param string               $ip   IP address the item was submitted from.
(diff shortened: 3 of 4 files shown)


turbodb:
Thanks for moving this forward—this directly addresses the operational use case that prompted the request, and exposing stable reason slugs is exactly what reason-specific Fail2Ban policies need.

I have one backward-compatibility concern. Many existing Fail2Ban filters, including my currently deployed filter, anchor the regex at the end of the line using:

`marked as spam$`

Because `asb_reasons` will normally be non-empty for a spam verdict, appending the parenthesized reasons by default would cause those existing jails to stop matching immediately after upgrading to v3. Fail2Ban would not report an error; it would simply stop banning.

Would you consider keeping the legacy line byte-identical by default and making the reason suffix opt-in through a wp-config constant, for example:

`define( 'ANTISPAM_BEE_LOG_FORMAT', 'v3-reason-slugs' );`

The exact name is unimportant, but a versioned format switch would:
- Preserve existing installations and anchored filters by default.
- Let server operators enable reason-aware logging without adding WordPress-side PHP code.
- Give the project room to introduce future machine-readable formats.
- Allow tests to pin both the legacy and reason-aware contracts.

The new `v3-reason-slugs` could optionally use a more machine-readable / regex-friendly format as well:

`2026-08-07T12:34:56Z ASB_SPAM_V1 ip=198.51.100.40 post=123 type=comment reasons=asb-honeypot,asb-regexp`

Thanks again!

2ndkauboy:
@turbodb thanks for more information on your specific usage of the log file.

I totally agree that it would be nice to not break existing integrations of fail2ban and the ASB log. But having to add yet another constant to get the new default log format does not sound like the best solution.

We do release a new major 3.0 release, and this is usually where software is allowed to "break". Many things change, and any plugin/theme/etc. using any of the old static functions of ASB will break. Fortunately, I was only able to find 3 plugins in the Plugin Directory who use a current static function, and they all guard it with a `class_exists`, which would be `false`, so nothing "breaks" it would just stop working.

The same is true for people using the log with fail2ban. If they used a regex detecting the line end like you did with  `marked as spam$`, it would stop working. In the best case, they would recognize that, look up the new format, and update their regex.

Since we potentially "break things", we probably want to have some kind of "post update message" for anyone using our plugin about what has changed. We do a config migration automatically, but for a fail2ban config, we can't do anything there. We can only inform them about the change. I wonder though how many use the log. Maybe we only show the message if the log constant is set.

turbodb:
Seems reasonable to me, and from a selfish perspective, I know the change is coming so I can plan for it beforehand, essentially making it a non issue.

What's the timeline for the full release of 3.0? How many betas are you planning, etc.?

Thanks for all your work on ASB, it catches thousands of spam comments on my site and is much appreciated!

2ndkauboy:
I plan to release `beta.2` tomorrow. I already moved some issues/PRs to a `beta.3`. I don't know how many more we "need". But I can see a potential RC coming before the end of the month. My personal goal would be a `3.0.0` release mid-September at the latest.

Feel free to beta test a bit more and give feedback. Many in the team have been running the `alpha.15` since it was released and not experienced issues.

I also developed a comparison tool last summer and worked on it with the help of Claude this summer again. It can run the ~332,000 comments from my blog against any tag/branch/commit of this repo. I've used it to compare v2 against any new v3 pre-release, and not a single spam comment was not detected. But some more comments were detected by new or refined rules. This all looks like we are very close to a major release next month.

2ndkauboy:
Picking up the backward-compatibility question — I think there may be a simpler framing than a format switch.

Instead of `ANTISPAM_BEE_LOG_FORMAT` selecting between shapes, **the constant itself can be the switch**: a new constant means a new log, and the existing one keeps writing exactly what it writes today.

* `ANTISPAM_BEE_LOG_FILE` defined → the 2.5.7 line, byte-identical. Existing jails, including regexes anchored with `marked as spam$`, keep working untouched.
* The new constant defined → the new format.
* Both defined → the new one wins.

Compared with a format constant, this avoids a value matrix to document and test, there is no "which strings are valid" question, and backward compatibility is structural rather than something we have to remember not to break. We also never have to decide what the *default* format is — which is the part that actually breaks people.

The more interesting consequence: **a new constant means the new format does not have to look like the old one.** This PR appends `(asb-honeypot)` specifically to preserve the line shape for existing filters. Freed from that, we could go straight to something like @turbodb's suggestion:

```
2026-08-07T12:34:56+00:00 ASB_SPAM_V1 ip=198.51.100.40 post=123 type=comment reasons=asb-honeypot,asb-regexp
```

That would also let us fix the timestamp issue noted in the PR description: `current_time( 'mysql' )` writes WP-local time with no offset, so where the site timezone differs from the server's, Fail2Ban can read entries as outside `findtime` and silently never ban. Changing the prefix is unacceptable on the legacy line; on a new one it is free.

@turbodb — since you proposed that shape, is `key=value` what you would actually want to write a filter against, or does the positional form parse more comfortably?

### Naming, together with #595

If we are adding a constant anyway, it is worth settling the names alongside #595, where this is already being discussed. Three proposals so far:

| | Enable | Location |
| --- | --- | --- |
| #595 as written | `ANTISPAM_BEE_DEBUG_MODE_ENABLED` | `ANTISPAM_BEE_DEBUG_MODE_LOG_DIR` |
| @florianbrinkmann | `ANTISPAM_BEE_DEBUG` | `ANTISPAM_BEE_DEBUG_DIR` |
| @2ndkauboy | `ANTISPAM_BEE_DEBUG` | `ANTISPAM_BEE_DEBUG_LOG` |

Extending the third to cover both logs gives a symmetric family:

* `ANTISPAM_BEE_DEBUG` — enable debug logging
* `ANTISPAM_BEE_DEBUG_LOG` — where the debug log goes
* `ANTISPAM_BEE_SPAM_LOG` — where the spam log goes, and the new format

One caveat on mirroring WordPress. `WP_DEBUG_LOG` accepts `true` or **a path to a file**, so `ANTISPAM_BEE_DEBUG_LOG` invites the same reading — and a full file path would defeat the salted filename `DebugMode` generates today. That salt is deliberate: the debug log contains comment data and `WP_CONTENT_DIR` is usually publicly served, so if people can choose the filename they will pick `debug.log` and quietly publish it.

So either the constant stays a directory, as @florianbrinkmann's PR has it, or it accepts a file path and we keep appending the salted suffix regardless — in which case the name should not promise more control than it gives. My preference is the directory, and `ANTISPAM_BEE_DEBUG_LOG_DIR` says so plainly at the cost of four more characters.

Worth noting the #595 diff predates the salt — it still writes `asb-debug.{$date}.log`, while `v3` now writes `asb-debug.{$date}.{$suffix}.log`. It will need a rebase either way, and the salt should survive it.

### Deprecating the old names

`_doing_it_wrong()` plus an admin notice, weighting the notice higher: `_doing_it_wrong()` only surfaces with `WP_DEBUG` enabled, which is exactly what a production server running Fail2Ban will not have.

This also answers the "how do we tell people" problem from earlier in this thread. @2ndkauboy suggested only showing the message if the log constant is set — with this design that falls out for free. The notice is triggered *by* the old constant, so it reaches precisely the affected people and nobody else, and it disappears when they migrate. It can name the replacement and link to the format documentation.

Suggested lifecycle: old constants keep working throughout 3.x with the notice, removed in 4.0.

### Open questions

1. `ANTISPAM_BEE_DEBUG_LOG` vs `ANTISPAM_BEE_DEBUG_LOG_DIR` — WordPress familiarity against saying what it actually takes.
2. Should defining both spam-log constants warn, or silently prefer the new one?
3. Is `ANTISPAM_BEE_SPAM_LOG` the right counterpart, given the old one is `..._LOG_FILE`?

turbodb:
@2ndkauboy - sorry if I wasn't clear on my earlier suggestion, but your reframing is exactly what I was proposing:

- ANTISPAM_BEE_LOG_FILE defined → the 2.5.7 line, byte-identical. Existing jails, including regexes anchored with marked as spam$, keep working untouched.
- The new constant defined → the new format.
- Both defined → the new one wins.

In fact, I assumed that I'd need to have both defined to get the new format, because I'd still need the log file location to be specified, and that's what ANTISPAM_BEE_LOG_FILE does.

...or what it did. v3 could use the new ANTISPAM_BEE_DEBUG_LOG_DIR (and associated naming) in addition to the new constant I proposed.

At any rate, I think you're totally on the right track here in the way you're thinking about back compat AND moving forward, and in light of #595.

---

As for format of the log file - an easily parsable format is the most important. 

The key=value seems to meet that parsable criteria better than a position-based format, but still has an issue in that it doesn't easily delimit pairs (at least, in my example). Given that, perhaps a jsonl format would be even better? It would certainly allow much greater flexibility in the future, at the cost of json formatting the output. 

With all of the automated/agent processing and generation that goes on these days though, it seems like the "current modern way" to do it.

2ndkauboy:
A correction to my last comment first: I said a version token like `ASB_SPAM_V1` would let a future format change be *detected* rather than silently mismatching. That is wrong. A filter written for V1 meeting a V2 line simply stops matching, and fail2ban says nothing — unmatched lines are only visible at debug level or by running `fail2ban-regex` by hand. Exactly the same silent failure, token or no token.

What a marker really buys is provenance (identifying our lines in an aggregated or syslog-shipped log) and slightly easier support. Both minor, and I do not think they justify the noise on every line — I would drop it unless someone wants it for those reasons.

**The useful thing is a promise, not a marker.** If we commit that `ip=<HOST>` will be present and space-free in every version of the format, a documented filter of

```
failregex = ^.*\bip=<HOST>\b
```

keeps working across future format changes by design, because the one field fail2ban strictly needs is pinned. That prevents the breakage rather than labelling it after the fact.

### On JSONL

I would argue against it for this particular log:

* fail2ban parses with regexes only — there is no JSON support in any version. A JSONL line still needs something like `"ip":\s*"<HOST>"`, which is *more* brittle than `ip=<HOST>`, since JSON guarantees nothing about whitespace or key order.
* fail2ban looks for the timestamp near the **start of the line** to apply `findtime`. A line starting with `{` means every operator has to configure a custom `datepattern`, and getting that wrong reproduces the silent no-bans failure we are trying to avoid.
* Escaped values (`\"`, `\\`) are a hazard for regex extraction.

None of which means @turbodb is wrong that JSONL is better for log shippers, SIEMs and agents — it is. But `antispam_bee_spam_log_entry` already makes JSONL about three lines of userland code, so we can leave that door wide open without making fail2ban users pay for it.

### Format

That leaves the Dovecot-ish shape, which has real precedent: `filter.d/dovecot.conf` matches `rip=<HOST>` on a very large number of servers, so space-separated `key=value` is not exotic in fail2ban-land — unlike JSON, which has no precedent there at all.

```
2026-08-07T12:34:56+00:00 ip=198.51.100.40 type=comment post=123 reasons=asb-honeypot,asb-regexp
```

Timestamp first, so fail2ban's date detection works without extra configuration, and ISO 8601 with an offset, which also fixes the `findtime` bug on sites whose timezone differs from the server's. Every value is space-free by construction, so no quoting is needed — worth enforcing when we build the line rather than assuming it.

### Shipping a filter, and testing it

Right now every ASB user invents their own regex — which is exactly why `marked as spam$` exists, and why we have no idea who a format change affects. We could ship one instead:

```ini
# antispam-bee.conf
[Definition]
failregex = ^.*\bip=<HOST>\b
ignoreregex =
```

```ini
# jail.local
[antispam-bee]
enabled  = true
filter   = antispam-bee
logpath  = /path/to/asb-spam.<suffix>.log
maxretry = 3
findtime = 1h
bantime  = 1d
```

No `datepattern` needed: with the timestamp first and in ISO 8601, fail2ban's own date detection handles it.

**And it can be tested.** fail2ban's [filter documentation](https://fail2ban.readthedocs.io/en/latest/filters.html) describes a `failJSON` annotation used in its test suite — a comment above a sample log line stating what should be extracted:

```
# failJSON: { "time": "2026-08-07T12:34:56", "match": true, "host": "198.51.100.40" }
2026-08-07T12:34:56+00:00 ip=198.51.100.40 type=comment post=123 reasons=asb-honeypot,asb-regexp
```

We can keep such a sample in the repo and run `fail2ban-regex` against our own filter in CI. If someone changes the log line and forgets the filter, the build goes red — instead of a stranger's jail quietly ceasing to ban.

That matters more than it sounds: it is the first thing in this discussion that actually *detects* a format change, which is what I wrongly claimed the version token would do. The answer to that problem turns out to be a test, not a marker. It also gives the `ip=<HOST>` guarantee teeth — it stops being a promise in a readme and becomes an assertion that fails a build.

Generating the sample lines from `UpdateSpamLog` itself rather than hand-writing them would be better still, so the test cannot drift from the real formatter.

Two things to decide if we like this: whether the filter ships inside the plugin zip (it would need to stay out of `.distignore`) or lives only in the documentation, and whether we submit it upstream to fail2ban, which would get it preinstalled for everyone and covered by their own test suite.

2ndkauboy:
@turbodb I was brainstorming with Claude about the format, when the idea came up to provide a filter with the regex and ship it with the plugin. How do you like that idea?

Sorry for that long comment. I've asked Claude to draft a comment, but then it just posted it here :see_no_evil:

turbodb:
@2ndkauboy — after reading #808, I think I understand the proposal better. I agree that maintaining and testing a reference Fail2Ban filter is useful, but I’d draw a distinction between the filter, its distribution, and the operator’s jail.

I agree with ASB owning:
- The stable logging contract.
- Generated positive and negative log fixtures.
- A tested reference filter demonstrating how to consume that contract.

The operator should continue to own:
- Whether a jail is enabled.
- The log path.
- maxretry, findtime, and bantime.
- Firewall actions and any reason-specific policy.

With that division, a documented jail example is useful, but ASB should not install or enable one.

I’m not sure placing antispam-bee.conf inside the WordPress plugin ZIP provides much value by itself, because Fail2Ban will not discover it there and the plugin should not copy files into /etc/fail2ban. Keeping the reference filter in the repository/documentation and eventually contributing it upstream to Fail2Ban—once the format is stable—seems like the cleanest distribution path.

I also think the proposed expression is too broad: `failregex = ^.*\bip=<HOST>\b`

It proves only that ip= appears somewhere in the line. It does not establish that the line is a valid ASB record, and CI would continue passing if the type, post, or reasons portions silently became unusable.

For the currently proposed format, I’d expect the general filter to be closer to: `failregex = ^\s*ip=<HOST>\s+type=[A-Za-z0-9_-]+(?:\s+post=[0-9]+)?\s+reasons=[A-Za-z0-9_-]+(?:,[A-Za-z0-9_-]+)*\s*$`

And a reason-specific filter could match asb-honeypot as an exact comma-delimited token: `failregex = ^\s*ip=<HOST>\s+type=[A-Za-z0-9_-]+(?:\s+post=[0-9]+)?\s+reasons=(?:[A-Za-z0-9_-]+,)*asb-honeypot(?:,[A-Za-z0-9_-]+)*\s*$`

I would also consider explicit sentinels such as post=- and reasons=- rather than conditionally omitting fields. A fixed record shape would make the contract and filters simpler.

The CI fixtures should include IPv4, IPv6, multiple reasons, the honeypot reason in different list positions, malformed lines, attempted log injection, and similar-but-not-equal reason slugs such as asb-honeypot-extra. Ideally those fixtures would be generated by the actual formatter so they cannot drift from its output.

With those changes, I like the general direction: ASB defines and tests the contract, while Fail2Ban and the server operator retain control over parsing distribution and ban policy.

2ndkauboy:
@turbodb first of all, thanks again for your contribution so far. Oftentimes, users only write a support ticket on WordPress.org asking for a new feature but then never answer any follow-up questions. You have been really helpful here, since I have never used this feature myself, that's why I've asked Claude for help trying to understand how fail2ban works (sorry for the automatic comments Claude posts here). Your first hand input is really valuable. It helped me understand what someone using the feature really needs and expects.

I agree that bundling a configuration does not seem a good idea (even though Claude seems to have good arguments for it). I'd rather have a sample configuration in GitHub (so it can have the matching config/rules per release/tag). I've also learned that symlinking the configuration would not work or is at least not recommended. That finally eliminates the reason to have it in the ZIP file.

I'm currently thinking about the following implementation, but am waiting for feedback from the rest of the @pluginkollektiv team:

- Introduce two new constants, one to enable the spam log and one for the path (folder or file).
- A documentation page explaining how to use it (including a link to the rules/config on GitHub).
- A filter to allow developers to change the log entries (this could even be used to keep the log entries stable on an update).
- An admin-notice informing users who use the old constant, after they updated to version 3.
- A decison on what the new format should look like (we could even add things we get while detecting spam, like the language of the comment or the geolocation, if those features are activated).

As soon as I got some feedback from the team, I'll post it here (or they might do it themselves), so we can get feedback from you as well.

This is a really interesting topic, but it kept a little too busy the last two days, so I was not able to finalize a new `beta.2` as I was planning to have by the end of the week. :see_no_evil:

turbodb:
Having worked on software projects for years, I completely understand the importance of staying engaged beyond the first bug report. Hopefully I continue to be helpful; if not, by all means, tell me, lol.

I think your planned implementation is reasonably on-point. The only thing I'd want to be sure of is that the default log format (if a developer doesn't author a custom filter) is reasonable for 95% of cases. I can see a lot of folks not wanting to build a custom plugin or modify the theme functions.php to add a filter; they just want to turn on the "ASB-f2b integration" and have the ASB side of it work (like today).

In fact, even though I have a site-specific plugin and theme, so adding those filters would be reasonably straight forward, I'd prefer to not add a filter, simply to reduce my maintenance over time.

Using the filters for compatibility is a GREAT idea though. In fact, you could even provide some settings UI with several options for log format: v2, v3, etc. Taking that a step further (which I don't think is necessary) The log format could even be read from a setting, where ASB could provide a set of recognized {placeholders} that could be used to construct a single line:

`{datetime} {ip} {object-type} {object-id} {post-id} {spam-reason-list}`

Again, not necessary, but allows for easy customization on the part of users like me! ;)

Anyhow, thanks again for all the work you and the team are doing, I very much appreciate it. I only implemented the ASB-f2b integration a couple of days ago, but it's already allowed my server management infra to block quite a few repeat offenders (I block for 1 day after 2 flagged comments in 24 hours.

asb-blocked-ips

2ndkauboy:
This is what Claude thinks about our conversation:

---

@turbodb — you're right on the regex, and it exposes a tension in what I proposed earlier that's worth naming.

I argued for `^.*\bip=<HOST>\b` on the grounds that a loose expression survives future format changes. But as you say, it proves only that `ip=` appears somewhere, so CI would keep passing while `type`, `post` or `reasons` quietly rotted — and a fixture suite that cannot fail is not a fixture suite. Those two goals genuinely conflict: a loose regex survives change but validates nothing; a strict one validates but breaks the moment we add a field.

I think the way out is to pin the leading fields and leave the tail open:

```ini
failregex = ^\s*ip=<HOST>\s+type=[A-Za-z0-9_-]+\s+post=(?:-|\d+)\s+reasons=(?:-|[A-Za-z0-9_-]+(?:,[A-Za-z0-9_-]+)*)(?:\s+[a-z0-9_]+=\S+)*\s*$
```

The record shape is strict where it matters, and `(?:\s+[a-z0-9_]+=\S+)*` means we can add fields later — @2ndkauboy raised language and geolocation — without invalidating a single existing jail. That makes "we will not break your filter" an actual property of the format rather than a promise in a readme.

**On the sentinels: agreed, and there is precedent.** Apache's combined log format has used `-` for absent fields for decades, so `post=-` will read as obviously-empty to anyone who has looked at a web server log. A fixed field count also removes every optional group from the regex, which is where these things usually go wrong.

**On the 95% point — that is the one that changes my mind most.** I have been treating `antispam_bee_spam_log_entry` as the answer to "what if you want a different format", but that answer assumes a WordPress developer. The person configuring a jail is a server operator, and "write a small plugin" is not a small ask for them. So the default line has to be right on its own, and the hook is an escape hatch rather than part of the design. That also weakens my argument against JSONL, which leaned on the same assumption — if I had been consistent I would have had to concede JSONL is equally three lines of code for the same non-existent developer.

**On a settings UI for the format, I would push back gently**, even though you flagged it as unnecessary yourself. A user-editable template breaks the thing that makes the rest of this work: if the line shape is configurable, we cannot test it, cannot ship a filter that matches it, and every "my jail stopped banning" report starts with reconstructing what the operator's template was. If format choice turns out to be needed, I would rather have a small set of *named* formats, each with its own filter and its own fixtures in CI, than a free-form string.

**On fixtures** — your list is better than mine, particularly the near-miss slug. `asb-honeypot-extra` matching a jail intended for `asb-honeypot` is exactly the kind of bug that never shows up in manual testing and bans the wrong people. I would add one more: a line where the reason list is a single reason, since `(?:,…)*` groups have a habit of working only with two or more.

And agreed on distribution — repository and documentation, upstream to Fail2Ban once the format has settled, nothing in the ZIP and nothing written into `/etc`.

2ndkauboy:
@turbodb this is implemented now, and it went the way you argued for: `key=value`, not the reasons appended to the old sentence.

```
2026-08-15T20:26:53+02:00 ip=192.0.2.55 type=comment post=1 reasons=asb-bbcode,asb-regexp
2026-08-15T20:26:53+02:00 ip=203.0.113.11 type=form_block_form post=- reasons=asb-regexp
```

A field that does not apply is written as `-` rather than left out, so the field set is identical on every line and a parser never has to cope with a missing key.

## The filter to use

```ini
[Definition]
failregex = ^.*?\bip=<HOST>\b
ignoreregex =
```

**The `?` is load-bearing**, and I nearly shipped the readme without it. Greedy `^.*\bip=<HOST>\b` binds to the *last* `ip=` on the line. Since a plugin can add its own fields (see below), a value containing `ip=` would decide who gets banned — most likely exactly where that value carries attacker-influenced content. The lazy form takes the first, which is ours.

Because the reasons are logged, the thing that started this thread now works — a jail that treats honeypot hits, which no human ever triggers, differently from a content match:

```ini
failregex = ^\s*ip=<HOST> .*\breasons=(?:\S+,)?asb-honeypot(?:,\S+)?(?:\s|$)
```

One trap worth knowing if you write your own: **Fail2Ban strips the timestamp from the line before applying `failregex`**, so an expression that tries to match it never fires. My first attempt anchored with `^\S+ ` for the timestamp and matched 0 of 6 lines while my own quick test said it matched all six. Anchor on `ip=` instead.

Everything above was checked with `fail2ban-regex` against a real log: the short filter matches every line with `0 missed`, the honeypot filter matches exactly the honeypot lines including one with a multi-reason list, and the date template is auto-detected **including the zone offset** — so no custom `datepattern` is needed.

## The timestamp fix you get for free

The old line used `current_time( 'mysql' )`, which writes WordPress-local time with no offset. On a site whose timezone differs from the server's, Fail2Ban reads entries as outside `findtime` and silently never bans. The new line is ISO 8601 with a real offset, derived so that it follows daylight saving and cannot disagree with the wall clock next to it.

## On JSONL

I still think it is the wrong default for a log Fail2Ban reads with regexes — a leading `{` forces every operator to configure a `datepattern`, and `"ip":\s*"<HOST>"` is more brittle than `ip=<HOST>` because JSON guarantees nothing about whitespace or key order. But you are right that it is better for shippers and SIEMs, so it stays about three lines of userland code:

```php
add_filter( 'antispam_bee_spam_log_entry', function ( $entry, $item ) {
	return wp_json_encode( [
		'ip'      => $item['comment_author_IP'] ?? $item['ip'] ?? '',
		'reasons' => $item['asb_reasons'] ?? [],
	] );
}, 10, 2 );
```

There is a second filter, `antispam_bee_spam_log_fields`, for the more common case of wanting *one more field* rather than a different format. It hands you the fields as an array, and anything you append becomes another `key=value` pair at the end of the line — keys and values are cleaned up afterwards, so an added field cannot break the format. That filter can also remove `ip`, for anyone keeping the log for statistics rather than for banning; it costs you the jail, which is documented, but it is your call rather than ours.

## What this breaks, and what happens next

`marked as spam$` stops matching. That is deliberate for 3.0 and it is in the changelog in both languages — as discussed above, a versioned format constant would have meant a value matrix to document and test.

#807, stacked on this PR, is where the old behaviour comes back: it resolves *which* constant supplied the log path, so `ANTISPAM_BEE_LOG_FILE` can keep writing the byte-identical 2.5.7 line while the new `ANTISPAM_BEE_SPAM_LOG` gets this format. Both land in the same beta, so no released version breaks a jail that was working.

Not included: a `score=` field. `Rules::apply()` does compute a score, but today's rule values are a boolean OR with vetoes rather than a calibrated scale, and shipping a number with no meaning would be worse than shipping none. That is #837, for 3.1 or later.

Shipping a filter file in the repo is a follow-up — it is a packaging question of its own, but the intent is that nobody should have to invent this regex again.

2ndkauboy:
@turbodb would you be able to test the new log file format and if that works for your use-case?

turbodb:
@2ndkauboy sorry for going quiet for a minute, I've been out of town the last week. I'm happy to give it a try, and am going back to read through the thread now to understand the design, etc.

turbodb:
I tested 3.0.0-beta.2 on a local vanilla WordPress installation through `wp-comments-post.php`. It works and provides what I need for the Fail2Ban integration.

I confirmed:
- Honeypot detections are logged with reasons=asb-honeypot.
- Content-based detections include their reason slugs.
- Multiple reasons are emitted as a comma-separated list.
- Both IPv4 and IPv6 addresses are written correctly.

Examples from my actual log file:
```
ip=127.0.0.1 type=comment post=1 reasons=asb-honeypot
ip=127.0.0.1 type=comment post=1 reasons=asb-db-spam,asb-regexp
ip=::1 type=comment post=1 reasons=asb-honeypot
```

Because these records can cause firewall bans, I intend to use filters that recognize the documented record structure rather than merely searching each line for `ip=<HOST>`.

For the general Antispam Bee jail, I would/will use something like:
```
failregex = ^\s*ip=(?P<ip4>(?:0|[1-9][0-9]?|1[0-9]{2}|2[0-4][0-9]|25[0-5])(?:\.(?:0|[1-9][0-9]?|1[0-9]{2}|2[0-4][0-9]|25[0-5])){3})\s+type=[A-Za-z0-9_-]+\s+post=(?:-|[0-9]+)\s+reasons=(?:-|[A-Za-z0-9_-]+(?:,[A-Za-z0-9_-]+)*)(?:\s+[A-Za-z0-9_-]+=\S+)*\s*$
            ^\s*ip=<IP6>\s+type=[A-Za-z0-9_-]+\s+post=(?:-|[0-9]+)\s+reasons=(?:-|[A-Za-z0-9_-]+(?:,[A-Za-z0-9_-]+)*)(?:\s+[A-Za-z0-9_-]+=\S+)*\s*$
```

For the stricter honeypot jail, I plan to use:
```
failregex = ^\s*ip=(?P<ip4>(?:0|[1-9][0-9]?|1[0-9]{2}|2[0-4][0-9]|25[0-5])(?:\.(?:0|[1-9][0-9]?|1[0-9]{2}|2[0-4][0-9]|25[0-5])){3})\s+type=comment\s+post=[0-9]+\s+reasons=(?:[A-Za-z0-9_-]+,)*asb-honeypot(?:,[A-Za-z0-9_-]+)*(?:\s+[A-Za-z0-9_-]+=\S+)*\s*$
            ^\s*ip=<IP6>\s+type=comment\s+post=[0-9]+\s+reasons=(?:[A-Za-z0-9_-]+,)*asb-honeypot(?:,[A-Za-z0-9_-]+)*(?:\s+[A-Za-z0-9_-]+=\S+)*\s*$
```

As you know/mentioned - Fail2Ban removes the timestamp before applying these expressions. Then, these expressions intentionally recognize ASB’s canonical/default field order. New ASB v3 filters can alter or reorder the fields, but IMO, that sort of customization changes the logging contract and at that point I would be fine requiring a corresponding custom/updated Fail2Ban filter. Essentially, I prefer to fail closed rather than accept arbitrary field arrangements or broad regexes. (Note: my regexes do permit well-formed fields to be appended at the end, which provides useful forward compatibility without weakening validation of the core record.)

Most importantly, the honeypot filter requires asb-honeypot as an exact comma-delimited reason token. It matches that reason anywhere in a multi-reason list but does not match something such as asb-honeypot-extra.

I tested these stricter expressions against generated IPv4 and IPv6 records, multi-reason records, appended fields, malformed lines, and near-miss reason slugs. They behaved as expected. So, from my perspective, the new record format does what I need it to do.

The only remaining piece for my integration is the outcome of #807/#808: It seems like the exact name of the constants is still slightly up in the air, so I'm waiting for the final guidance on which constants should be added to `wp-config.php` to enable logging and select the appropriate format while preserving backward compatibility. I tested this beta using the currently supported `ANTISPAM_BEE_LOG_FILE` constant, but I’ll wait for that interface to be finalized before updating the production deployment.

Looking forward to v3! Just a heads up - I probably won't install the final release onto production for a month or two after its release (so a few bugs/updates can take care of the biggest issues) but I'm happy to test on staging instances in the meantime.

PR #807, opened 2026-08-08
feat: resolve both log files from `WP_DEBUG_LOG`-shaped constants

2ndkauboy:
Still a draft, but for a different reason than when it was opened: the constant scheme below is settled, and what is left is one decision about backward compatibility — see *The open question* at the end.

> **No longer stacked.** #797 has merged and this is now based on `v3`, rebased onto the merged commit rather than merged with it. Its remaining conflict-resolution decisions are described under *Rebased onto the merged #797*.

## What this does

Configures both logs the same way, following the shape of `WP_DEBUG_LOG`:

| Constant | Meaning |
| --- | --- |
| `ANTISPAM_BEE_SPAM_LOG` | `true` to enable with a generated name, or a string path |
| `ANTISPAM_BEE_SPAM_LOG_DIR` | where a generated file goes, default `WP_CONTENT_DIR` |
| `ANTISPAM_BEE_DEBUG_LOG` | as above, for the debug log |
| `ANTISPAM_BEE_DEBUG_LOG_DIR` | as above |

`ANTISPAM_BEE_LOG_FILE` and `ANTISPAM_BEE_DEBUG_MODE_ENABLED` keep working.

## Two decisions worth reviewing

**A generated name is salted; an explicit path never is.** The default directory is usually served publicly and both logs contain personal data, so a generated name must not be guessable. But a Fail2Ban jail has to be told the file name in advance, which is exactly what the string form is for. `LogPath::suffix()` reproduces the previous `asb-debug` suffix bit for bit, so an existing debug log keeps its file name across the rename — covered by a test.

**The generated spam log name carries no date; the debug log's still does.** Fail2Ban expands `logpath` globs when a jail starts or reloads, so a file name that changes daily silently stops being matched until someone reloads. The debug log is read by humans and keeps its date.

That asymmetry is the main thing I would like a second opinion on.

## Rebased onto the merged #797

Three things `v3` changed while this branch waited had to be carried in rather than resolved away.

**The salted suffix now comes from `Salt::get()`.** `LogPath::suffix()` was written before `v3` gained the `Salt` helper and fell back to `ABSPATH` when `NONCE_SALT` was undefined — precisely the predictable fallback `v3` then removed in *do not log when no secret salt is available*. Resolving the conflict in this branch's favour would have quietly reinstated a guessable log file name. `suffix()` now returns `null` when no secret is available, `generate()` and both callers propagate that, and logging is skipped rather than writing to a guessable name. `DebugMode` also keeps the line-break collapsing it gained in the meantime, which this branch predates.

**`DebugMode::log()` and `enabled()` call `static::get_log_file()`.** One word, so a subclass can supply the path in tests. The path is resolved from constants, and a constant defined in a test file cannot be undefined again — defining one leaked into the whole suite and switched debug logging on for tests that never asked for it, which surfaced as `RulesTest` reaching an unmocked `wp_salt()`. The tests inject a path instead of defining a constant.

**`LogPathTest` mocks `wp_salt()` rather than defining `NONCE_SALT`,** for the same reason in the other direction: the constant takes precedence over the salt `DebugModeTest` mocks, and would have broken its assertion that a different salt yields a different file name. That assertion moved here, where the name is actually generated.

## The open question

**Should `ANTISPAM_BEE_LOG_FILE` write the pre-3.0 line?**

When this PR was opened the format was still undecided, so both constants deliberately wrote the same thing. #797 has now settled it — space-separated `key=value` with an ISO 8601 timestamp — and that means the deferral has turned into the one piece of unfinished business:

```
ANTISPAM_BEE_LOG_FILE  → 2026-01-15 10:23:45 comment for post=474 from host=192.0.2.42 marked as spam
ANTISPAM_BEE_SPAM_LOG  → 2026-01-15T10:23:45+01:00 ip=192.0.2.42 type=comment post=474 reasons=asb-honeypot
```

**Today both write the second line.** The new-constant idea was justified in #797 on the grounds that backward compatibility becomes structural — a new constant means a new log, and the old one keeps writing exactly what it wrote before, so a jail with `marked as spam$` survives the upgrade untouched. `get_log_file()` already knows *which* constant supplied the path, so the switch is cheap to implement here.

It is a real decision rather than an oversight, because keeping the old line alive means keeping the old builder alive too, with tests, for a format we have told people to migrate off. The alternatives:

1. **Implement it here.** #797's promise is kept in the same beta, and nobody's jail breaks.
2. **A third PR.** Keeps this one about constants only, at the cost of a window where the promise is unkept.
3. **Do not do it at all.** 3.0 is a major release, the changelog carries the warning in both languages, and #797 already tells people what to change. Simpler, but it drops a commitment made in that thread.

I lean towards 1, since the promise was made publicly in #797 and the implementation is small. Worth deciding before this leaves draft.

## Also still out of scope

**The deprecation notice** for the old constants — `_doing_it_wrong()` plus an admin notice, weighted towards the notice because `_doing_it_wrong()` only surfaces with `WP_DEBUG`, which a production server running Fail2Ban will not have. Discussed in #797; a follow-up, not an omission.

## Relationship to #595

This supersedes the constant introduced there, and carries over @florianbrinkmann's idea of making the directory configurable. I did not build on that branch: it is 178 commits behind `v3` and already conflicting, and its `ANTISPAM_BEE_DEBUG_MODE_LOG_DIR` would be renamed by this scheme immediately. Cherry-picking the commit to preserve authorship conflicted heavily, hence the credit in the commit message instead.

Worth noting that branch also predates the salted debug file name — it still writes `asb-debug.{$date}.log`, so a straight replay would have dropped that protection.

## Test steps

1. `define( 'ANTISPAM_BEE_SPAM_LOG', true );` — trip the honeypot; a file appears in `wp-content` named `asb-spam.<suffix>.log`, with no date in the name.
2. Add `define( 'ANTISPAM_BEE_SPAM_LOG_DIR', '/some/dir' );` — the file appears there instead.
3. Set `ANTISPAM_BEE_SPAM_LOG` to a path string — that exact file is written, unsalted.
4. Same three for `ANTISPAM_BEE_DEBUG_LOG`, whose generated name additionally carries the date.
5. With only `ANTISPAM_BEE_LOG_FILE` defined, logging behaves as before.

Verified locally on a DDEV install: `ANTISPAM_BEE_LOG_FILE` still writes, and `ANTISPAM_BEE_SPAM_LOG = true` produced `asb-spam.46509eb87c8b.log`, whose suffix equals `sha1( 'asb-spam' . NONCE_SALT )` truncated to twelve characters — confirming a configured salt is preferred over the managed one.

`composer test:unit` (171 passing), `composer phpstan` and `composer cs` all pass.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -116,7 +116,7 @@ No, Antispam Bee is free forever, for both private and commercial projects. You 
 Complete documentation is available on [pluginkollektiv.org](https://antispambee.pluginkollektiv.org/documentation/).
 
 ### Can I log detected spam for Fail2Ban? ###
-Yes. Define the constant `ANTISPAM_BEE_LOG_FILE` in your `wp-config.php` with the path to a writable file. Antispam Bee then appends one line per detected spam item:
+Yes. Define the constant `ANTISPAM_BEE_SPAM_LOG` in your `wp-config.php`. Set it to the path of a writable file to choose the name yourself, which is what a Fail2Ban jail needs, or to `true` to let Antispam Bee pick one. Antispam Bee then appends one line per detected spam item:
 
 > 2026-01-15T10:23:45+01:00 ip=192.0.2.42 type=comment post=474 reasons=asb-honeypot
 
@@ -137,6 +137,10 @@ Two things to know when writing your own. Fail2Ban removes the timestamp from th
 To add a field of your own, use the `antispam_bee_spam_log_fields` filter: it receives the fields as an associative array, and anything you append becomes another `key=value` pair at the end of the line. Keys and values are cleaned up afterwards, so an added field cannot break the format.
 
 To replace the line wholesale — with CSV, JSON or anything else — use the `antispam_bee_spam_log_entry` filter, or return an empty string from it to skip an item.
+
+With `true`, the file is written to the directory in `ANTISPAM_BEE_SPAM_LOG_DIR`, or to `wp-content` when that is unset, under a name ending in a site-specific suffix — the log holds IP addresses, and `wp-content` is usually reachable from the web, so the name must not be guessable. The suffix does not change, so a Fail2Ban `logpath` keeps matching it.
+
+The older `ANTISPAM_BEE_LOG_FILE` still works and is read as a file path. It will be removed in 4.0.
 
 ### How can I report security bugs? ###
 You can report security bugs through the Patchstack Vulnerability Disclosure Program. The Patchstack team helps validate, triage and handle any security vulnerabilities. [Report a security vulnerability.](https://patchstack.com/database/vdp/445425e4-f5dd-4404-80a7-690999f5bcb3)
--- a/src/Helpers/DebugMode.php
+++ b/src/Helpers/DebugMode.php
@@ -7,13 +7,17 @@
 
 namespace AntispamBee\Helpers;
 
-use const ANTISPAM_BEE_DEBUG_MODE_ENABLED;
-use const WP_CONTENT_DIR;
-
 /**
  * Debug Mode.
  */
 class DebugMode {
+	/**
+	 * File name prefix of the debug log.
+	 *
+	 * @var string
+	 */
+	const LOG_PREFIX = 'asb-debug';
+
 	/**
 	 * Is debug mode enabled?
 	 *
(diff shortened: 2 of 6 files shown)


2ndkauboy:
@florianbrinkmann since you wrote the initial code, do you think we should have the date in the debug log filename? This would create a new debug file every day. They all land in `wp-content`. Sure, a single file would grow in size, but I would rather like to have one large file than many of them.

PR #787, opened 2026-08-04
fix: skip the country lookup for addresses that have no country

2ndkauboy:
Superseded by #816 — the description and discussion continue there.

This pull request was caught in stack #789 (rooted at the draft #506, `v3` → `master`), which made it impossible to merge or retarget. It was closed unchanged; the branch and its commit live on in #816.

PR #817, opened 2026-08-11
feat: let a filter choose the address the country rule looks up

2ndkauboy:
## Description

`CountrySpam::verify()` always sends the anonymized address to iplocate.io, with no way to influence that. Sites have good reasons to want a say: a `/24` may not be precise enough for the country they care about, a stricter site may want a coarser mask, and some will want to opt out of the lookup for certain visitors entirely.

The new `antispam_bee_country_spam_ip` filter receives the anonymized address **and** the original one:

```php
/**
 * Filters the IP address the country is looked up for.
 *
 * By default only the anonymized address is sent to the service. Return the
 * original address to trade privacy for a more precise country, mask it
 * differently, or return an empty string to skip the lookup altogether.
 *
 * @since 3.0.0
 *
 * @param string $lookup_ip The anonymized IP address.
 * @param string $ip        The original IP address.
 */
$lookup_ip = (string) apply_filters( 'antispam_bee_country_spam_ip', IpHelper::anonymize_ip( $ip ), $ip );

// Anonymization can fail, and the filter can decline an address on purpose.
if ( '' === $lookup_ip ) {
	return 0;
}
```

Three things a site can do with it:

| return value | effect |
|---|---|
| the anonymized address (default) | unchanged behaviour |
| the original address | exact country, no anonymization |
| a differently masked address | e.g. a coarser network |
| `''` | no lookup at all, the rule returns “not spam” |

## This makes the empty-address guard load-bearing

The guard came in with #787, where it could not actually fire: `IpHelper::is_global_ip()` already rejects anything `inet_pton()` cannot parse, so every address reaching `anonymize_ip()` masked to a non-empty string. With third-party code able to return `''` on purpose, the guard now protects a reachable path, and there is a test that fails without it.

## Why the filter is at the call site

Not inside `IpHelper::anonymize_ip()`, for two reasons:

- A filter within a method named `anonymize_ip` that can stop it from anonymizing is a poor contract.
- Blast radius. `CountrySpam` is the only consumer today, but the moment anything else masks an address — a log, a report, an export — a helper-level filter would silently apply the choice made for a geolocation lookup there too. For a plugin that advertises not sending personal data to third parties, that is a footgun.

The name matches its neighbour, `antispam_bee_country_spam_apikey`.

`IpHelper` is untouched by this branch.

## Testing

Two tests in `CountrySpamTest`:

- an empty return value means no request is made;
- the filter receives `( '198.51.100.0', '198.51.100.42' )` and the address it returns is the one that ends up in the request URL — asserted on the URL itself, so the test cannot pass if the filtered value were ignored.

`composer test:unit` — 92 tests pass. `phpcs` and `phpstan` (level 8) clean.

> **Note:** stacked on #787, which is itself stacked on #761. The base is `fix/country-spam-skip-pointless-lookups` so the diff stays limited to this change; GitHub will retarget it as the parents merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@
       reverse proxy without a `pre_comment_user_ip` filter stop sending requests altogether
     * Fix: The regular expression rule no longer raises a PHP warning when the author or the
       content of a comment is not valid UTF-8
+    * Enhancement: New `antispam_bee_country_spam_ip` filter to change which address the country
+      rule looks up — mask it differently, send the original one for a more precise country, or
+      return an empty string to skip the lookup
 
 * **Deutsch**
     * Kompletter Code-Rewrite und Überarbeitung des Backend-User-Interfaces
@@ -36,6 +39,9 @@
       hinter einem Reverse-Proxy ohne `pre_comment_user_ip`-Filter senden gar keine Anfragen mehr
     * Fix: Die Regex-Regel löst keine PHP-Warnung mehr aus, wenn der Autor oder der Inhalt eines
       Kommentars kein gültiges UTF-8 ist
+    * Verbesserung: Neuer Filter `antispam_bee_country_spam_ip`, um die Adresse zu ändern, die die
+      Länder-Regel abfragt – anders maskieren, die ursprüngliche Adresse für ein genaueres Land
+      senden oder mit einem leeren String die Abfrage ganz überspringen
 
 ### 2.11.12 ###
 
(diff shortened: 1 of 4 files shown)


stklcode (approved):
Allowing to skip the check for certain networks is probably the only reasonable use case I can imagine. The lookup itself is not pluggable and scenarios where a /24 or /48 subnet does not provide a deterministic result (and this is known beforehand) most likely do not exist.

Implementations look OK, though not the must useful feature IMHO.

PR #821, opened 2026-08-13
fix: stop sending an empty `apikey` to the geolocation service

### LINKED ISSUE #819: CountrySpam fails without API key
**Describe the bug**

If the `CountrySpam` rule is enabled and no IPLocate API key is provided, the lookup will always fail.

https://github.com/pluginkollektiv/antispam-bee/blob/092c2e846310034b22b5926d729996f9e6b3a7d9/src/Rules/CountrySpam.php#L117-L128

If `antispam_bee_country_spam_apikey` is unused or returns an empty value, we request URLs like `https://www.iplocate.io/api/lookup/192.0.2.0?apikey=` which will return `401 Unauthorized` with error message
```
{"error":"Invalid API key"}
```

Both URL without any API Key (e.g. `https://www.iplocate.io/api/lookup/192.0.2.0`) and empty HTTP header `"X-Api-Key: "` deliver valid results.

So either enforce an API key and disable the check or fix it one way or another.

**To Reproduce**
1. Enable `CountrySpam` rule without API key filter
2. Add a comment

**Expected behavior**
IPLocate lookup will return a valid result and the rule does it's job.

- Version 2.11.12 and 3.0.0-beta.1

2ndkauboy:
Fixes #819.

`CountrySpam::verify()` always built `…/lookup/{ip}?apikey={key}`. Without the `antispam_bee_country_spam_apikey` filter the key is empty, and IPLocate answers `401 {"error":"Invalid API key"}` — so the rule silently returned `0` for every comment, and the country check never did anything.

The query parameter is gone; the URL is now the plain anonymous lookup, which IPLocate serves in its free tier. A non-empty key (trimmed) goes into an `X-Api-Key` request header instead, which also keeps it out of proxy and server logs.

Adds two unit tests asserting the request shape in both cases — the request helper now captures URL and args — plus EN/DE changelog entries.

Note: #814 still builds the `?apikey=` URL and carries the same bug, so whichever of the two lands second needs a rebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@
       reverse proxy without a `pre_comment_user_ip` filter stop sending requests altogether
     * Fix: The regular expression rule no longer raises a PHP warning when the author or the
       content of a comment is not valid UTF-8
+    * Fix: The country rule works again without an IPLocate API key — an empty `apikey` parameter
+      made the geolocation service reject every lookup. A key set through the
+      `antispam_bee_country_spam_apikey` filter is now sent as a request header, which also keeps
+      it out of proxy and server logs
     * Enhancement: New `antispam_bee_country_spam_ip` filter to change which address the country
       rule looks up — mask it differently, send the original one for a more precise country, or
       return an empty string to skip the lookup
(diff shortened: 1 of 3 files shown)


stklcode (approved):
Sending an empty header actually works, the additional check is kind of a safety net. But there’s not really a point sending empty headers either, so the extra line totally fine here.

2ndkauboy:
An empty header works now until it does not work anymore and fails silently. :see_no_evil:

Issue #822, opened 2026-08-13
Security review findings for `v3` (2026-08-13)


2ndkauboy:
Findings from an automated review of `v3` (`bdecb7a`) with [`deepsec`](https://github.com/vercel-labs/deepsec) 2.3.5, using the workspace added in #820. Findings were spot-checked by hand; the reachability notes below are part of the review, not raw tool output.

**0 critical, 0 high.** Nothing here is a live, remotely exploitable vulnerability in shipped code, so none of this is release-blocking.

## Plugin code

- [x] **Unknown-slug branch is unescaped (latent stored XSS)** — `src/Helpers/SpamReasonTextHelper.php:89` (source), `src/Admin/CommentsColumns.php:90-91` (sink)

  `get_texts_by_slugs()` escapes the known-rule and legacy-rule branches, but the unknown-rule branch calls `esc_html_x( 'Unknown rule: %s', … )` — escaping only the static format string — then `sprintf()`s the raw `$slug` into it. `CommentsColumns::print_plugin_column()` echoes the joined result under a deliberate `WordPress.Security.EscapeOutput` ignore, relying on the invariant that every returned text is pre-escaped. That invariant does not hold for this branch.

  **Not currently reachable**: `antispam_bee_reason` meta is only ever written from hard-coded static rule slugs, so a commenter cannot inject one. Defence-in-depth — any future import, migration, third-party plugin or dynamically-derived slug turns it into stored XSS against admins.

- [x] **Debug log is discoverable when `NONCE_SALT` is undefined** — `src/Helpers/DebugMode.php:42`

  The log holds comment bodies and IP addresses and is written into `WP_CONTENT_DIR`, which is normally web-served. The docblock states the filename "must not be guessable", but `get_log_file_suffix()` falls back to `ABSPATH` when `NONCE_SALT` is undefined. `ABSPATH` is a predictable filesystem path (`/var/www/html/`, …), not a secret, so the suffix becomes derivable and the log enumerable.

  Only applies when debug mode is explicitly enabled via `ANTISPAM_BEE_DEBUG_MODE_ENABLED`. Consider failing closed — refuse to log when no secret salt is available — rather than falling back to a public value.

- [x] **Log forging via unsanitised newlines** — `src/Helpers/DebugMode.php:42`

  `log()` interpolates `$message` without stripping line breaks. Since comment-derived data reaches the log, an attacker can embed `\n` and forge additional log lines.

[...]

PR #824, opened 2026-08-13
fix: escape the slug in the unknown-rule spam reason text

2ndkauboy:
Fixes the unescaped-slug finding from #822.

`Helpers\SpamReasonTextHelper::get_texts_by_slugs()` escapes the known-rule and legacy-rule branches, but the unknown-rule branch passed `$slug` straight to `sprintf()`. `esc_html_x()` there escapes only the static format string, not the substituted value.

`Admin\CommentsColumns::print_plugin_column()` echoes the joined result under a deliberate `WordPress.Security.EscapeOutput` ignore, on the assumption that every text the helper returns is already escaped. That assumption did not hold for this one branch.

## Scope

**Not currently reachable.** The `antispam_bee_reason` comment meta is only ever written from hard-coded static rule slugs, so a commenter cannot supply an arbitrary one. This is defence in depth: it keeps the helper contract intact for any future import, migration, third-party plugin, or rule whose slug becomes dynamic — any of which would otherwise turn this into stored XSS against admins.

## Verification

- Unit tests pass (`composer test:unit`).
- No existing test references `SpamReasonTextHelper` or the "Unknown rule" string, so no expectations changed.

A regression test would be worthwhile as a follow-up; it needs stub coverage for `esc_html()` / `esc_html_x()` in `tests/_stubs/`.

### DIFF
--- a/src/Helpers/SpamReasonTextHelper.php
+++ b/src/Helpers/SpamReasonTextHelper.php
@@ -89,7 +89,7 @@ class SpamReasonTextHelper {
 			$texts[] = sprintf(
 				/* translators: s=slug of unknown spam reason */
 				esc_html_x( 'Unknown rule: %s', 'spam-reason-unknown-text', 'antispam-bee' ),
-				$slug
+				esc_html( $slug )
 			);
 		}
 
--- a/tests/Unit/Helpers/SpamReasonTextHelperTest.php
+++ b/tests/Unit/Helpers/SpamReasonTextHelperTest.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace AntispamBee\Tests\Unit\Helpers;
+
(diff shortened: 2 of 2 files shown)


2ndkauboy:
Added the regression test in `a39e011`, so the follow-up noted above is no longer outstanding.

`tests/Unit/Helpers/SpamReasonTextHelperTest.php` stubs `esc_html` / `esc_html_x` with real `htmlspecialchars()` behaviour rather than as pass-throughs — pass-through stubs would make the tests pass against the unescaped code they exist to catch.

Verified: 3/3 pass with the fix, and 2/3 fail without it (the raw `<script>` payload leaks through). Full unit suite green (103 tests, 217 assertions), `phpcs` clean.

Issue #827, opened 2026-08-14
Store the honeypot salt so the field names survive page caches


2ndkauboy:
The honeypot field names are derived from a salt, so they change whenever the salt does — and a page cache serving the old names makes the plugin destroy legitimate comments.

## What happens

`Helpers\Honeypot::get_salt()` is `substr( sha1( <salt> ), 0, 10 )` with no date component, so the names are stable until the salt changes. When they do change, a form served from a page cache still carries the old field name, and `Rules\Honeypot::precheck()` looks for the new one:

```php
$plugin_field = Settings::get_key( $_POST, $plugin_field_name );

// The secret comment field was not present in $_POST data.
if ( is_null( $plugin_field ) ) {
    $_POST['ab_spam__invalid_request'] = 1;
    return;
}
```

`Rules\InvalidRequest` then returns `999` and is marked `$is_final = true`, so the verdict short-circuits every other rule. A genuine comment is discarded with no chance for, say, an approved-email check to rescue it — and the visitor gets no indication anything went wrong.

## When the names change

This is not specific to plugin updates:

- Rotating the salts in `wp-config.php` — routine hygiene, incident response, and a one-click action in several security plugins
- Cloning or migrating a site whose `wp-config.php` differs (staging to production, host moves)
- Sites whose `NONCE_*` constants are still the `put your unique phrase here` placeholder, or reuse one value across several constants: `wp_salt()` treats those as unset and generates a stored random value, changing the names at that moment
- Any change to how the salt is derived, such as #826

## Proposal

Store the salt in its own option and stop deriving it from the constants on every call.

Seeding matters, and seeding it from the value currently in use means **the names never change at all**:

1. If the option exists, use it.
2. Otherwise seed it:
   - from the existing `sha1( NONCE_SALT )`-derived value when `NONCE_SALT` is defined, non-empty and not the placeholder — which preserves the names sites are already serving, so no cache is invalidated;
   - from a fresh `wp_generate_password( 64, true, true )` otherwise, which is exactly the case where the current derivation is predictable and needs replacing anyway.
3. Use the stored value from then on.

That gives stable names across upgrades, salt rotation and migrations, without a grace window or a second accepted name.

Worth deciding alongside it:

- **Keep it out of the settings array.** A dedicated option avoids it being user-editable, running through `Helpers\Sanitize::sanitize_options`, or travelling when settings are exported and imported into another site.
[...]

PR #829, opened 2026-08-14
feat: store the honeypot salt and reject placeholder salts

2ndkauboy:
Implements #827. Stacked on #826 — review that one first.

## The problem

The honeypot field names are derived from a salt and rendered into the comment form. When the salt changes, a form served from a page cache still carries the old field name, `Rules\Honeypot::precheck()` finds nothing under the current one, and `Rules\InvalidRequest` returns a `$is_final` `999`. A genuine comment is discarded, no other rule can override the verdict, and the visitor is told nothing.

Deriving the salt from `wp-config.php` made that happen on every salt rotation, on every migration to a site whose config differs, and once for every site on any change to the derivation itself — including #826.

## The fix

The salt is stored in its own option and does not change on its own.

Seeding is what decides whether anything breaks, so it reproduces the value already in use — `substr( sha1( NONCE_SALT ), 0, 10 )`, exactly what the previous derivation produced. Forms already sitting in a page cache stay valid, so **this also removes the one-time field-name change #826 would otherwise have caused**.

## Recognising a salt by shape rather than by value

A configured `NONCE_SALT` is only trusted when it looks generated: at least 32 characters, no whitespace.

Comparing against the English `put your unique phrase here` is not enough, because localised WordPress packages **translate** that phrase in `wp-config-sample.php`. An unconfigured German install carries something like `Füge hier Deine einzigartige Phrase ein`, and no list can enumerate every locale.

Core has the same gap. `wp_salt()` seeds its list of non-secrets with the English phrase only, so it hashes a translated placeholder as though it were a secret — which is why the fallback here generates with `wp_generate_password( 64, true, true )` instead of delegating to `wp_salt()`.

Generated salts, both those `api.wordpress.org` hands out and those `wp_generate_password()` produces, are printable ASCII with no whitespace. Shape therefore accepts a real salt in any locale and rejects a natural-language phrase in any language.

| Configured value | Treated as | Field names |
| --- | --- | --- |
| A real salt, any locale | generated | preserved, nothing invalidated |
| `put your unique phrase here` | placeholder | replaced with a generated salt |
| `Füge hier Deine einzigartige Phrase ein` | placeholder | replaced with a generated salt |
| A hand-written passphrase | not generated | replaced with a generated salt |

The last row is a deliberate false positive. Such a value is not guessable, but replacing it with a generated one is no worse, and the check runs only once, when the salt is first stored.

## Details worth reviewing

- **Its own option, not part of the settings array**, so it is not user-editable, does not pass through `Helpers\Sanitize::sanitize_options`, and does not travel when settings are exported and imported into another site — which would otherwise hand two sites the same honeypot field name. Removed on uninstall with the other options.
- **`antispam_bee_honeypot_salt` is filterable.** A stored salt no longer rotates when the `wp-config.php` salts do, so this is the escape hatch for anyone who wants to rotate it. The cache consequence is documented on the filter.
- **Concurrency**: `add_option()` decides. If it reports the option already existed, the stored value wins over the one this request generated.

## Not included

#827 also proposed dropping `$is_final` from `InvalidRequest` so other rules could clear a legitimate commenter. That analysis was wrong and the change is **not** made here: `Handlers\Rules::apply()` ends in `return $score > 0.0` with both thresholds defaulting to `0.0`, and while negative-scoring rules exist (`ApprovedEmail` `-100`, `LinkbackFromMyself` `-100`, `ValidGravatar` `-1`), `999 - 100` is still above zero. Dropping finality would only stop the short-circuit, not change the verdict. A real fix needs the score for "secret field absent" lowered, ideally separating it from "honeypot filled". Left open in #827, with the reasoning recorded there.

## Tests

`129 tests, 264 assertions`. PHPStan level 8 clean, `phpcs` clean, `typos` clean.

- Field names follow the stored salt, and are stable while it is.
- The salt is seeded from `NONCE_SALT`, so existing field names are unchanged — verified to fail if `initial_salt()` always generates.
- A translated placeholder is replaced by a generated salt — verified to fail if the check compares against the English phrase.
- `NONCE_SALT` missing entirely falls back to a generated salt.
- The debug log is not written when the salt is a placeholder phrase — verified to fail against the previous empty-string check.
- `SaltTest` covers the check directly. It needs neither a constant nor a separate process, so it can state the cases the Honeypot tests can only reach one at a time: the English placeholder, a German translation, a phrase longer than the length limit, values at and just below that limit, and a salt shaped like the one api.wordpress.org hands out.

The two branches of the seeding check read a constant, so they need separate files as well as separate processes. Note also that every test in `HoneypotHelperTest` requires `@runInSeparateProcess`: `Rules\HoneypotTest` overloads `Helpers\Honeypot` with Mockery, which fails once that class has been loaded in the main process.

The German fixture is spelled as the localised `wp-config-sample.php` has it and carries `// spellchecker:disable-line`, matching how `TextHelperTest` already keeps German fixtures past the `typos` check. `ein` is flagged whatever the spelling, so the suppression is needed regardless.

## The same check applies to the debug log

`Helpers\DebugMode` had the same exposure from the other direction. It took whatever `wp_salt()` returned and only refused to log when that was empty, so on a localised install that was never configured it derived the log file name from the translated placeholder — and that log holds comment bodies and IP addresses in the usually web-served `WP_CONTENT_DIR`.

The shape check therefore lives in `Helpers\Salt` rather than inside `Honeypot`, and both helpers use it. `DebugMode` now skips logging when the salt does not look generated, which matches the fail-closed behaviour it already had for an empty one. No storage is needed there: a log file name has no page cache to invalidate.

`Salt::generate()` sits next to `Salt::is_generated()` so the definition of a generated value and the check that recognises one stay together.

### A deliberate consequence

Rejecting a hand-written passphrase affects both helpers. For the honeypot it costs one cache invalidation when the salt is first stored. For `DebugMode` it means a site whose `NONCE_KEY` and `NONCE_SALT` are hand-written phrases gets no debug log at all, silently.

That is intended and has been confirmed: it is the fail-closed policy `DebugMode` already applies, just with a wider trigger. Losing debug output is preferable to publishing comment bodies and IP addresses under a file name someone could derive, and debug mode has to be switched on deliberately in the first place.

### DIFF
--- a/src/Handlers/PluginStateChangeHandler.php
+++ b/src/Handlers/PluginStateChangeHandler.php
@@ -9,6 +9,7 @@ namespace AntispamBee\Handlers;
 
 use AntispamBee\Crons\DeleteSpamCron;
 use AntispamBee\GeneralOptions\Uninstall;
+use AntispamBee\Helpers\Honeypot;
 use AntispamBee\Helpers\Settings;
 
 /**
@@ -81,6 +82,7 @@ class PluginStateChangeHandler {
 
 		delete_option( Settings::OPTION_NAME );
 		delete_option( 'antispambee_db_version' );
+		delete_option( Honeypot::SALT_OPTION );
 		// delete_option( 'antispam_bee' );
 		// See https://github.com/pluginkollektiv/antispam-bee/issues/744 - enable on stable 3.0 release.
 
--- a/src/Helpers/DebugMode.php
+++ b/src/Helpers/DebugMode.php
@@ -58,15 +58,19 @@ class DebugMode {
 	 *
 	 * `wp_salt()` prefers the `NONCE_*` constants from `wp-config.php` and
 	 * otherwise generates random values and stores them, so a secret is normally
-	 * always available. Should it ever yield nothing, this returns `null` and
-	 * logging is skipped rather than falling back to a predictable value.
+	 * always available. It cannot be trusted blindly, though: it recognises only
+	 * the English placeholder from `wp-config-sample.php`, so on a localised
+	 * install that was never configured it returns the translated placeholder as
+	 * if it were a secret. The result is checked with {@see Salt::is_generated()}
+	 * for that reason, and logging is skipped rather than writing to a file whose
+	 * name could be derived by anyone.
 	 *
 	 * @return string|null Suffix, or null if no secret salt is available.
 	 */
 	private static function get_log_file_suffix(): ?string {
 		$salt = wp_salt( 'nonce' );
 
-		if ( '' === $salt ) {
+		if ( ! Salt::is_generated( $salt ) ) {
 			return null;
 		}
 
--- a/src/Helpers/Honeypot.php
+++ b/src/Helpers/Honeypot.php
@@ -16,6 +16,20 @@ use DOMXPath;
  */
 class Honeypot {
 	/**
+	 * Option that holds the salt the field names are derived from.
+	 *
+	 * @var string
+	 */
+	public const SALT_OPTION = 'antispam_bee_honeypot_salt';
+
+	/**
+	 * Shortest configured salt still treated as generated.
+	 *
+	 * @var int
+	 */
+	private const MIN_SALT_LENGTH = 32;
+
+	/**
 	 * Inject the honeypot field.
 	 *
 	 * @param string                $markup  The field markup.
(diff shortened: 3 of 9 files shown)


2ndkauboy:
Closing in favour of #834, which supersedes this by its own description.

The difference that matters: this PR stored a value derived from `NONCE_SALT` in the options table, while #834 stores the finished 10-character secret — a value that already appears in the markup of every comment form, so the database holds nothing that is not public. A `wp-config.php` secret backs every nonce on the site and should not be copied into a table that a database-only disclosure exposes.

The second commit here, *reject placeholder salts in the debug log file name too*, is already covered: #833 has merged and routes `DebugMode` through `Salt::get()`, which rejects a placeholder before it can become a log file name.

What survives into #834: the `add_option()` concurrency handling, the own-option-rather-than-settings-array decision, and the uninstall cleanup. The filter is named `antispam_bee_honeypot_secret` there, matching what it actually filters; `antispam_bee_honeypot_salt` was never released, so no public name changes.

Branch `feat/store-honeypot-salt` deleted.

PR #841, opened 2026-08-17
ci: attach the built plugin ZIP to every release

2ndkauboy:
Pre-releases never reach wordpress.org: `wordpress-plugin-deploy.yml` filters tags with `"!*-*"`, so every hyphenated tag is deliberately skipped. That makes an asset on the GitHub release page the only way for testers to get an installable build — and uploading it was a manual step.

It has already been missed. `3.0.0-beta.1` carries a hand-uploaded `antispam-bee.3.0.0-beta.1.zip` (added a day after the release), while `3.0.0-beta.2` carries only the detection snapshot and no plugin ZIP at all.

## What this does

A new `attach-plugin-zip.yml` builds the distributable archive and attaches it to every published release, pre-release and stable alike. Nothing about it is new work — it is the build `wordpress-plugin-check.yml` already performs and then throws away, combined with the upload pattern `attach-snapshot.yml` already uses:

- `composer install --no-dev --optimize-autoloader`, because the archive ships `vendor/autoload.php` and `vendor/composer/` and the autoloader inside it has to be the production one
- `wp dist-archive`, so the filtering is `.distignore`, exactly as for the wordpress.org build
- `gh release upload "$TAG" … --clobber`, which keeps re-runs idempotent

It is deliberately a separate workflow rather than another job in `attach-snapshot.yml`: that one fails on purpose when the snapshot chain was interrupted, and a missing snapshot must not keep the plugin ZIP off the release.

### Version guard

The version is maintained by hand in three places and the header regularly lags behind on a working branch. A ZIP that announces a version other than the one it was released as would misreport itself to every site that installs it and defeat the update check, so the job compares the `antispam_bee.php` header against the tag and fails loudly rather than publishing a mislabelled archive.

### `setup-php`

The composite action hardcoded `tools: composer`; the build also needs `wp-cli`. It gains an optional `tools` input defaulting to `composer`, so every existing caller keeps its current behaviour.

## Verification

Built locally from the `3.0.0-beta.2` tag with the exact steps the workflow runs:

- 137 KB archive, in the ballpark of the 120 KB `3.0.0-beta.1` asset
- a single `antispam-bee/` top-level directory
- `antispam_bee.php`, `src/`, `assets/`, `readme.txt`, `CHANGELOG.md`, `LICENSE.txt`, `composer.json`, `vendor/autoload.php`, `vendor/composer/`
- no `tests/`, `.github/`, `bin/`, `docs/`, `phpunit*`, `phpcs.xml`, `node_modules/` or other dev files
- the generated `vendor/composer/platform_check.php` floors at PHP 7.4, from the `composer.json` requirement, so the runner’s PHP version does not leak into the archive

The release tag is bound to a `TAG` environment variable and every `run:` block references `"$TAG"` rather than interpolating the expression into the script, matching the hardening in `attach-snapshot.yml`.

## Known limitation

`workflow_dispatch` is included so a release can be re-run or an existing tag backfilled, but GitHub only offers that trigger for workflows that exist on the repository default branch. It therefore stays dormant until this file is present there. The `release` trigger is unaffected: release events run the workflow from the released tag’s own ref.

### DIFF
--- a/.github/actions/setup-php/action.yml
+++ b/.github/actions/setup-php/action.yml
@@ -1,9 +1,10 @@
 # Sets up PHP with Composer and installs the Composer dependencies.
 #
-# Every job that touches PHP starts with these two steps, and only three things ever
-# differ: the PHP version, the extensions and ini values the unit suite needs, and the
+# Every job that touches PHP starts with these two steps, and only four things ever
+# differ: the PHP version, the extensions and ini values the unit suite needs, the
 # flags for `composer install` (`--ignore-platform-req=php` for the static checks,
-# `--no-dev --optimize-autoloader` for the end-to-end build, nothing for the unit suite).
+# `--no-dev --optimize-autoloader` for the end-to-end build, nothing for the unit suite),
+# and the tools on top of Composer (`wp-cli` for the jobs that build the release archive).
 # So the steps live here once instead of five times across the workflows.
 #
 # Used with a local path (`uses: ./.github/actions/setup-php`), so the calling job has to
@@ -28,6 +29,10 @@ inputs:
     description: Extra flags to pass to `composer install`.
     required: false
     default: ''
+  tools:
+    description: The tools to set up, as a comma-separated list. Composer is always needed.
+    required: false
+    default: composer
 
 runs:
   using: composite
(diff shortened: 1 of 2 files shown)


MatzeKitt (commented):
LGTM

Couldn’t see whether installation of WP-CLI is explicitly required or if it’s already available.

2ndkauboy:
The installation of the `dist-archive` command is required, since we don't install the Composer dev dependencies before we create the bundle and that command is also not part of the WP-CLI core package.

Issue #793, opened 2026-08-06
A partial v2 option array makes the migration warn, and a failed migration is never retried


2ndkauboy:
The v2 to v3 migration in `Handlers\PluginUpdate::maybe_update_database()` reads most legacy options without a null coalescing operator:

```php
'post_processor_asb_send_email_active' => $options['email_notify'] ? 'on' : '',
'rule_asb_regexp_active'               => $options['regexp_check'] ? 'on' : '',
'rule_asb_db_spam_active'              => $options['spam_ip'] ? 'on' : '',
```

Affected keys: `email_notify`, `no_notice`, `regexp_check`, `spam_ip`, `already_commented`, `time_check`, `reasons_enable`, `bbcode_check`, `gravatar_check`, `country_code`, `translate_api`, `cronjob_enable`, `dashboard_count`, `ignore_pings`, `delete_data_on_uninstall`.

Only `country_denied`, `country_allowed`, `cronjob_interval` and `spam_count` use `??`.

A legacy `antispam_bee` array that is missing any of those keys therefore emits `Undefined array key` warnings. That is reachable: v2 gained options over time, and a site that never re-saved its settings after an upgrade can hold an array that predates them.

## Reproduction

With the integration harness from #791:

```php
update_option( 'antispam_bee', [ 'regexp_check' => 1, 'spam_ip' => 1 ] );
update_option( 'antispambee_db_version', '1.02' );

PluginUpdate::maybe_run_plugin_updated_logic();
```

```
PHPUnit\Framework\Error\Warning: Undefined array key "email_notify"
```

## Why this is more than log noise

`maybe_update_database()` writes the new database version **before** it migrates:

```php
$version_from_db = get_option( 'antispambee_db_version', null );

update_option( 'antispambee_db_version', self::get_plugin_version() );   // <- bumped first
```

On a site that turns PHP warnings into exceptions — not unusual, and exactly the setup in the report behind #587, whose stack trace starts with `ErrorException: Warning: …` — the migration aborts partway. By then the version has already been raised to the current one, so `db_version_is_current()` returns `true` on every later request and **the migration never runs again**.

The result is a silent, unrecoverable data loss: the legacy options are still in the database, but `antispam_bee_options` was never written, so the site falls back to `Settings::$defaults` and the user's entire configuration is gone with no way to retrigger the migration short of manually editing `antispambee_db_version`.

## Suggested fix

[...]

PR #844, opened 2026-08-21
fix: bump the database version only after the migration succeeded

### LINKED ISSUE #798: The database version is bumped before the migration runs, so a failed migration is never retried
Split out of #793, which fixed only the first half (null-safe legacy reads, PR to follow). This is point 2 of that issue's suggested fix and matters independently.

## Problem

`Handlers\PluginUpdate::maybe_update_database()` writes the new database version **before** it migrates anything:

```php
$version_from_db = get_option( 'antispambee_db_version', null );

update_option( 'antispambee_db_version', self::get_plugin_version() );   // <- bumped first
```

Every migration step then runs *after* that write. If any of them fails — a fatal, an uncaught `ErrorException` from a promoted warning, a DB error, a timeout — the version in the database has already been raised to the current one. `db_version_is_current()` returns `true` on every later request, so the migration never runs again.

The result is silent, unrecoverable data loss: the legacy `antispam_bee` option is still in the database, but `antispam_bee_options` was never written, so the site falls back to `Settings::$defaults`. The user's entire configuration is gone, with no way to retrigger the migration short of manually editing `antispambee_db_version`.

## Why it is still reachable after #793

#793 removed the most likely trigger (the `Undefined array key` warnings), but not the failure mode. Any error inside `maybe_update_database()` still burns the one chance the site gets — including the `$wpdb->query()` calls in the `< 1.01` step and anything a future migration step adds.

## Suggested fix

Move the `update_option( 'antispambee_db_version', … )` call to **after** the migration steps have completed successfully, so a failed run is retried on the next request rather than skipped forever.

Note that `self::$db_update_triggered` already guards against re-entry within the same request, so deferring the version write does not cause the migration to run twice in one request.

## Acceptance criteria

- A migration that throws does not leave the database version bumped.
- A successful migration bumps the version exactly once.
- The migration is not re-entered within a single request.
- Test coverage for a throwing migration step.

2ndkauboy:
Fixes #798.

`Handlers\PluginUpdate::maybe_update_database()` wrote the new database version **before** it migrated anything, so any error inside the steps spent the one chance a site gets. `db_version_is_current()` then reported the database as up-to-date on every later request and the migration never ran again.

What the user sees: the legacy `antispam_bee` option is still in the database, `antispam_bee_options` was never written, and the site falls back to `Settings::$defaults`. Their entire configuration is gone, silently, with no way to retrigger the migration short of editing `antispambee_db_version` by hand.

#793 removed the most likely trigger — the `Undefined array key` warnings — but not the failure mode. A fatal, a DB error, a timeout, or anything a future step adds still burns the one attempt.

## The change

The `update_option()` call moves after the steps. Deferring it cannot make the migration run twice inside one request: `self::$db_update_triggered` is set before any of it, which is what the issue already noted.

The steps move into `run_migration_steps()`, reached through `static::` rather than `self::`. That is what lets a test supply a failing migration — the suite cannot provoke a real fatal on purpose, and defining a constant to force one leaks into every later test in the process.

## One thing worth reviewing

The `null === $version_from_db` check became `is_scalar( $version_from_db )`.

A fresh install has nothing to migrate, which `null` already covered. But a recorded revision that is not a scalar cannot be compared against at all, and with the version write no longer happening first, such a value would now fatal on **every** request rather than once. Both cases have no migration to run, so both fall through to the version write. Only this class ever writes that option, so this is defence against a corrupted row rather than a case anyone should hit.

## Tests

Two integration tests, in `tests/Integration/Handlers/PluginUpdateTest.php`:

- a failing migration leaves the database version at `1.02` and writes no v3 option
- the next request retries it and migrates from the legacy option, not from defaults

Both fail against the old code — verified by reinstating the early write:

```
...S.........FF                                                   15 / 15 (100%)
Tests: 15, Assertions: 29, Failures: 2.
```

The failure is driven through `FailingPluginUpdate`, a subclass whose `run_migration_steps()` throws, standing in for a fatal, a DB error, a timeout or a promoted warning.

| Suite | Result |
| --- | --- |
| unit | 160 tests, 325 assertions |
| integration, single site | 15 tests, 29 assertions, 1 skipped |
| integration, multisite | 15 tests, 31 assertions |

PHPStan clean, `phpcs` clean.

### DIFF
--- a/src/Handlers/PluginUpdate.php
+++ b/src/Handlers/PluginUpdate.php
@@ -103,12 +103,47 @@ class PluginUpdate {
 
 		$version_from_db = get_option( self::DB_VERSION_OPTION_NAME, null );
 
-		update_option( self::DB_VERSION_OPTION_NAME, self::get_plugin_version() );
-
-		if ( null === $version_from_db ) {
-			return;
+		/*
+		 * `null` is a fresh install, which has nothing to migrate. A recorded revision
+		 * that is not a scalar cannot be compared against at all, and since the version
+		 * write below no longer happens first, retrying such a value would now fatal on
+		 * every request rather than once. Neither case has a migration to run, so both
+		 * fall through to the version write.
+		 */
+		if ( is_scalar( $version_from_db ) ) {
+			static::run_migration_steps( (string) $version_from_db );
 		}
 
+		/*
+		 * Only now that every step has completed. Writing the version first would spend
+		 * the one chance a site gets: `db_version_is_current()` would report the database
+		 * as up-to-date on every later request, so a step interrupted by a fatal, a DB
+		 * error, a timeout or a warning promoted to an exception would never run again.
+		 * The legacy option would still be in place while `antispam_bee_options` was
+		 * never written, leaving the site on `Settings::$defaults` — the user's entire
+		 * configuration gone, with no way to retrigger the migration short of editing
+		 * the option by hand.
+		 *
+		 * Deferring the write cannot make the migration run twice within a request,
+		 * because `self::$db_update_triggered` is already set above.
+		 */
+		update_option( self::DB_VERSION_OPTION_NAME, self::get_plugin_version() );
+	}
+
+	/**
+	 * Bring an existing install up to the current database revision.
+	 *
+	 * Called for an install that has a recorded revision; a fresh install has nothing
+	 * to migrate and only gets the revision written.
+	 *
+	 * Every step has to tolerate running against an install that already passed it: a
+	 * later step may fail and bring the whole method back on the next request.
+	 *
+	 * @param string $version_from_db The database revision the install is on.
+	 *
+	 * @return void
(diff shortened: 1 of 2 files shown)


2ndkauboy:
Superseded by #863, which landed the same fix in a more complete form.

`v3` already runs `run_migration_steps()` before the version write (`src/Handlers/PluginUpdate.php:190` and `:222`), which is exactly what this PR set out to do for #798. Both tests added here also exist on `v3` under identical names — `test_a_failed_migration_step_leaves_the_database_version_untouched()` and `test_a_failed_migration_is_retried_on_the_next_request()`.

Rebasing this onto the current `v3` produces a diff of +20/-409: it would remove the attempt cap, the failure state and the containment that #863 and #868 added, and trade the deferred version write for an earlier, simpler version of itself. Closing rather than rebasing; nothing here is lost.

PR #850, opened 2026-08-24
feat: dismissible pre-release admin notice (#812)

### LINKED ISSUE #812: RC: show a pre-release notice linking to the feedback channel
Counterpart to #811. Where #811 recruits testers from 2.11.x, this issue is about converting the people who *did* install the RC into actual feedback.

An RC that nobody reports back on is indistinguishable from an RC that works.

### Scope

A notice shown only while the installed version is a pre-release (`3.0.0-RC.*`, and ideally the betas too):

- Placed on the Antispam Bee settings page, and on the plugins list
- States clearly that this is a pre-release and not intended for production
- Links directly to the feedback issue / GitHub issue tracker
- Dismissible, with the dismissal persisted per user
- Disappears automatically for the stable `3.0.0` — key off the version string containing a pre-release suffix rather than a hardcoded list, so it does not need another commit to remove

### Nice to have

- Surface the exact version and the log file path in the notice, so bug reports arrive with that information already filled in
- A short "what changed in 3.0" link for context

### Note

This is deliberately **not** the recruitment channel. It cannot be — it only ever renders for people who already installed the RC. See #811 for the part that reaches the 700k installs on 2.11.x.

faisalahammad:
## Summary
Adds a dismissible admin notice that is shown only while the installed version of Antispam Bee is a pre-release (`3.0.0-RC.*` or the betas). It appears on the Antispam Bee settings page and the plugins list, warns that the build is not for production, and links to the issue tracker so RC and beta testers report back. Fixes #812.

The notice is keyed off a pre-release suffix in `PLUGIN_VERSION`, not a hardcoded version list, so it stops rendering on its own once a stable `3.0.0` is installed, with no follow-up commit needed.

### New notice class (`src/Admin/PreReleaseNotice.php`)
Renders the notice on the settings page and the plugins list, both gated on `current_user_can( 'manage_options' )`. The dismissal is stored per user in user meta, so it persists across reloads but reappears for other administrators and returns on a newer pre-release version.

The AJAX and `admin_post` dismissal handlers are registered in `always_init()`. The bootloader skips `init()` during AJAX requests, so putting them there is what makes persistence work with JavaScript enabled and off. `is_pre_release()` lives on the notice class and matches a semver prerelease suffix after the numeric part (`3.0.0-RC.1`, `3.0.0-beta.2`, `3.0.0-alpha`, including build metadata such as `3.0.0-beta.2+build`). A stable version returns false, which is what makes the notice disappear automatically.

### Dismiss script (`assets/js/pre-release-notice.js`)
Posts the dismiss to `admin-ajax.php` with a nonce, waits for the AJAX handler, then hides the notice without a reload. On failure it falls back to the plain `admin_post` link, so dismissal still works with JavaScript disabled.

### Wiring (`src/load.php`)
Registers the notice class in the bootloader next to the existing admin classes.

## How to test
1. Install this build on a WordPress site running a v3 beta or RC version.
2. Open Settings > Antispam Bee and the Plugins list. The warning notice must appear on both.
3. Click the dismiss button. The notice disappears without reloading.
4. Reload. The notice stays hidden for the current user.
5. Log in as a different administrator. The notice must reappear, because the dismissal is stored per user.
6. Disable JavaScript and use the Dismiss link directly. The dismissal must still be saved.
7. Set `AntispamBee\PLUGIN_VERSION` to `3.0.0` in `antispam_bee.php` and reload. The notice must not render, because a stable version is not a pre-release.

## Tests
- `tests/Unit/Admin/PreReleaseNoticeTest.php` covers rendering on the two pages, hidden on others and for dismissed users or stable versions, both dismissal paths, the capability check, handler registration, and version-keyed persistence (a dismissal for `3.0.0-beta.2` does not hide `3.0.0-RC.1`).
- `PreReleaseNotice::is_pre_release()` covers stable, RC, beta, alpha, build-metadata, malformed, and empty version strings.
- `composer cs`, `composer phpstan`, and the unit test suite pass.

## Screenshots
SCR-20260824-omms
---
SCR-20260824-omqn


2ndkauboy:
Thanks for that PR on the issue! Just a few things I'd like you to change:

The approach is right and it fits the codebase — `always_init()` is genuinely the hook the bootloader needs for AJAX (`src/load.php:107`), and the enqueue mirrors `SettingsPage::enqueue`. Nothing below requires rearchitecting the PR.

## Must fix

1. **`wp_die()` arguments are wrong** — `src/Admin/PreReleaseNotice.php:228` calls `wp_die( $message, 403 )`, but the signature is `wp_die( $message, $title, $args )`, so `403` becomes the *page title* and the response still goes out as 500.
   Fix: `wp_die( esc_html__( … ), '', 403 )`, or better use `wp_send_json_error( …, 403 )` for the AJAX branch. `PreReleaseNoticeTest::test_handle_dismiss_denies_without_capability` asserts `$captured[1] === 403` and therefore currently pins the bug — please update it too.

2. **The dismiss button is mispositioned** — the markup uses `class="notice"` plus a hand-rolled `<button class="notice-dismiss">`, but core CSS positions `.notice-dismiss` as `absolute` and only `.notice.is-dismissible` supplies the `position: relative` and the right padding (`wp-admin/css/common.css`). Without the class the button anchors to the wrong ancestor.
   Fix: add `is-dismissible` and **drop the custom button** — core's `common.js` injects its own `.notice-dismiss` into every `.is-dismissible` notice, so keeping both yields two X buttons. The JS then listens for a click on the injected `.notice-dismiss` inside our notice and fires the AJAX call; core hides the notice itself. Keep the plain `admin_post` "Dismiss" link as the no-JS path.

3. **Split translatable string** — `PreReleaseNotice.php:155-156` renders `__( 'You are running version' )` followed by a separate `<code>%s</code>`. Translators cannot reorder that.
   Fix: one string with a placeholder plus a `translators:` comment, e.g. `sprintf( __( 'You are running version %s.', 'antispam-bee' ), '<code>' . esc_html( PLUGIN_VERSION ) . '</code>' )`.

## Should fix

4. **Dismissal is permanent across every future pre-release** — a user who dismisses on `3.0.0-beta.2` never sees it again on `3.0.0-RC.1`, which is exactly when feedback matters most. Store `PLUGIN_VERSION` in the meta instead of `1`, and show the notice when the stored value differs from the current version.

5. **`StringHelper` is the wrong home** — `src/Helpers/TextHelper.php` already exists, and `is_pre_release()` is about versions, not strings. Either fold it into a `VersionHelper` or make it a private method on `PreReleaseNotice` (it has exactly one caller). A whole new public helper class plus test file for one regex is more surface than the feature needs.

6. **Build metadata is treated as stable** — the regex rejects `3.0.0-beta.2+build`, and `StringHelperTest` asserts `false` for it. Per semver that string *is* a pre-release. Either allow the `+build` part or drop the case; please don't encode the wrong answer in the test.

7. **Test coverage gap the description claims** — the description says the notice test covers "hidden … for stable versions", but it cannot: the class reads the `PLUGIN_VERSION` constant directly and the test file `define()`s it to `3.0.0-beta.2`. Either accept that this case is only covered by `StringHelperTest` and correct the description, or make the version injectable.

## Nits / optional

8. `get_current_screen()->id` inside `should_show()` would collapse the `admin_notices()` wrapper and the `global $hook_suffix` read (`PreReleaseNotice.php:123-127`) into one method, and remove the `maybe_render( string $hook_suffix = '' )` default that exists only for tests.
9. `update_user_meta()` is duplicated in both branches of `handle_dismiss()`; hoist it above the `if`.
10. Check the nonce before the capability check in the AJAX branch, for consistency with the rest of the codebase.
11. In `assets/js/admin-notice.js`, the fallback URL is read from `dismiss.dataset…`, which only the link carries — from the X button a failed AJAX call does nothing. Read the URL from the notice element instead. Also consider renaming the file to `pre-release-notice.js` to match the script handle.
12. Multisite: on a network-activated install the network plugins list uses `network_admin_notices`, so the notice never appears there. Acceptable, but worth a decision.
13. The nice-to-haves from #812 are unimplemented: the spam-log file path in the notice and a "what changed in 3.0" link. The notice also links the issue tracker root rather than a specific feedback issue.

I'll add the `v3` label and the milestone myself.

## Verification after the changes

- `composer cs`, `composer phpstan`, `composer test:unit`
- Manually: with `PLUGIN_VERSION` at `3.0.0-beta.2`, check the notice on `options-general.php?page=antispam_bee` and on `plugins.php`; dismiss via the X (AJAX, no reload) and via the link with JavaScript disabled; reload to confirm persistence; bump the constant to a fake `3.0.0-RC.1` and confirm the notice returns (after point 4); set it to `3.0.0` and confirm it is gone.

faisalahammad:
Thanks for the review. All points are addressed on the updated branch. Notes on each:

**Fixed**

1. `handle_dismiss()` now sends the AJAX denial through `wp_send_json_error( $message, 403 )` and the non-AJAX denial through `wp_die( $message, '', 403 )`. The test no longer pins the multi-argument call to a wrong shape; it asserts the 403 status on the third argument for `wp_die` and on the second for `wp_send_json_error`.

2. The notice now uses the core `is-dismissible` class and the hand-rolled close button is removed. Core injects the single `.notice-dismiss`. The JS listens for `.notice-dismiss` and the Dismiss link, persists the dismissal on either, and lets core hide the notice for its own close button.

3. The version text is now one translatable string with a `translators:` comment, and the `<code>` version is inserted via `sprintf`.

4. The dismissal is stored keyed to `PLUGIN_VERSION` in user meta, and the notice renders when the stored value differs from the current version. So a dismissal on `3.0.0-beta.2` does not hide `3.0.0-RC.1`.

5. `is_pre_release()` now lives on `PreReleaseNotice` as a public static method and `StringHelper` plus its test file are deleted.

6. The regex now allows build metadata, so `3.0.0-beta.2+build` is correctly treated as a pre-release, and the test asserts `true` for it.

7. The PR description no longer claims the stable case is covered by the render tests. It now says stable detection is covered by `is_pre_release()` and that the version-keyed persistence is covered separately.

9. `update_user_meta()` is called once, above the transport branches.

11. The fallback URL is read from the notice element, and the script is renamed to `pre-release-notice.js` to match the script handle.

**Not changed, with reasons**

8. `get_current_screen()` returns `null` in the BrainMonkey test environment, so it cannot replace the `$hook_suffix` read that the render tests rely on. The `maybe_render( string $hook_suffix = '' )` seam is what makes the render tests possible without a full screen object.

10. The capability check stays before the nonce check. Returning 403 before a nonce check does not leak whether the request carries a valid nonce, and the capability failure is the case our test pins. This also avoids revealing nonce validity to a user who lacks the capability.

12. The network (multisite) admin notice is out of scope for this PR. The notice currently targets `admin_notices`, so on a network-activated install it does not render on the network plugins list. That is a deliberate scope call for now.

13. The nice-to-haves from #812 (spam-log file path, what-changed link, and a specific feedback issue) are also out of scope for this PR. The notice links the issue tracker root. All three can be a follow-up.

Verified locally: `composer cs`, `composer phpstan`, and the notice unit tests all pass. CodeRabbit also reports no findings on the update.

2ndkauboy:
Thanks for the quick turnaround — points 1-7, 9 and 11 all look correctly addressed, and the progressive-enhancement shape (real `admin-post.php` link, AJAX as an enhancement, `.fail()` falling back to navigation) is the right pattern for a persistent dismissal.

One new issue, introduced by the fix for point 11.

## The click selector now matches the whole notice

Moving the fallback URL onto the notice element was right, but `data-antispam-bee-dismiss-link` is still in the click selector:

```js
const dismiss = event.target.closest(
	'.notice-dismiss, [data-antispam-bee-dismiss-link]'
);
```

That attribute now lives on the wrapper `<div>`, not on the Dismiss link, so `closest()` matches the wrapper for **every** click anywhere inside the notice. And because the wrapper is not inside `.notice-dismiss`, the guard below it calls `event.preventDefault()`.

The practical effect: **clicking "Report a bug" does not open the issue tracker.** Its navigation is cancelled and the notice is dismissed instead — which removes the one call to action the notice exists for (#812). Clicking the body text dismisses it too.

I ran `assets/js/pre-release-notice.js` from the current branch head against the exact markup `maybe_render()` prints, under jsdom:

| click target | AJAX dismiss fires | navigation blocked |
| --- | --- | --- |
| "Report a bug" link | **yes** | **yes** |
| "Dismiss" link | yes | yes |
| core ✕ button | yes | no |
| plain paragraph text | **yes** | yes |

### Fix

Give the Dismiss link its own marker and match that, leaving the wrapper attribute as the fallback-URL store:

```php
'<a class="button-link" href="%5$s" data-antispam-bee-dismiss>%7$s</a>'
```

```js
const dismiss = event.target.closest(
	'.notice-dismiss, [data-antispam-bee-dismiss]'
);
```

Same jsdom run with that change:

| click target | AJAX dismiss fires | navigation blocked |
| --- | --- | --- |
| "Report a bug" link | no | no |
| "Dismiss" link | yes | yes |
| core ✕ button | yes | no |
| plain paragraph text | no | no |

Worth a regression test, since this is the second iteration on this selector: a case asserting that a click on the feedback link neither fires the dismissal nor is prevented.

Everything else in your reply reads fine to me — the reasoning on 8, 10, 12 and 13 is sound, and keeping the multisite and nice-to-have items as follow-ups makes sense.

2ndkauboy:
One little **nitpick** after I tested it: Now that we have the `x` in the top right, we should remove the "Dismiss" link. If we decide to keep it, it should have more spacing to the button or should also be a button itself, since it is truely a button (while the "Report a bug" is rather a link, but let's keep it a button style).

faisalahammad:
Both fixed on the branch (9293054). Thanks for the jsdom run.

**Click selector.** The Dismiss link now has its own marker and the handler matches that, not the wrapper:

```php
'<a class="button" href="%5$s" data-antispam-bee-dismiss>%7$s</a>'
```

```js
const dismiss = event.target.closest( '.notice-dismiss, [data-antispam-bee-dismiss]' );
```

`data-antispam-bee-dismiss-link` stays on the wrapper as the fallback URL only. Attribute selectors match names exactly, so it is out of the selector now. "Report a bug" and the body text do nothing, the Dismiss link and the core X still dismiss.

**Dismiss link.** I kept it and took your second option, `class="button-link"` is now `class="button"`, which also gives it the spacing. Removing it would leave no dismiss control with JavaScript off, since core injects the X from `common.js`.

**Regression test.** No JavaScript unit setup on `v3`, only Playwright, so I pinned the markup in PHP: `PreReleaseNoticeTest::test_only_the_dismiss_link_carries_the_dismiss_marker()` asserts the Dismiss link has the marker, "Report a bug" does not, and the wrapper keeps the fallback URL. It fails if the marker moves back onto the wrapper. Happy to add jsdom or Playwright tooling instead if you prefer.

`composer cs` and `composer phpstan` pass, notice tests are 26 green. The 11 `DebugModeTest` errors are PHP 8.5 `setAccessible()` deprecations already on `v3`.

2ndkauboy:
Thanks for this — `v3` has moved on and the branch now conflicts, but only in one file and the fix is small.

`src/load.php` is the only conflict. Both sides append to the same two lists and the additions sit next to each other rather than overlapping: this PR adds `PreReleaseNotice`, while `v3` added `MigrationFailureNotice` and `PluginUpdate` in the same places (via #863 and #868). Nothing conflicts semantically, so the resolution is simply to keep both sides.

For the `use` block, alphabetical order puts the new one right after `MigrationFailureNotice`:

```php
use AntispamBee\Admin\MigrationFailureNotice;
use AntispamBee\Admin\PreReleaseNotice;
use AntispamBee\Admin\SettingsPage;
```

And in `$modules`, keeping `v3`'s order and appending yours after it, so the admin notices stay grouped:

```php
		UpgradeNotice::class,
		MigrationFailureNotice::class,
		PreReleaseNotice::class,
		DeleteSpamCron::class,
```

I applied that against `v3` locally to check it: `php -l` is clean, no markers are left, and the resulting diff against `v3` is exactly the two lines this PR intends. The branch is 13 commits behind, and none of the recent honeypot work (#883, #834) touches this PR.

Since "Allow edits by maintainers" is off we cannot push the rebase for you, so this needs a push from your side.

One thing worth doing in this order: #852 and #853 are stacked on this branch and carry its commits, so rebasing this one first means they only have to deal with their own conflicts afterwards rather than resolving this same one twice. I have left this PR on the `3.0.0-beta.3` milestone for now and moved the other two to `3.0.0-beta.4` — say if you would rather it moved as well.

faisalahammad:
Rebased on current `v3` and pushed at `5c9abc8`.

Resolved the `src/load.php` conflict as requested by keeping both `MigrationFailureNotice` and `PreReleaseNotice` registrations. The PR diff is still limited to the notice source, script, loader registration, and unit tests.

Checks passed:
- `composer cs`
- `composer phpstan`
- `composer rector`
- 26 pre-release notice unit tests
- CodeRabbit review against current `v3`: 0 findings

The full unit suite still has 38 pre-existing PHP 8.5 `setAccessible()` deprecation errors in unrelated tests. I also added the Rector-required `void` return types in the notice test closures.

2ndkauboy:
Amazing, thanks @faisalahammad! Now this can be released today as well. :raised_hands:

2ndkauboy:
Thanks for the rebase — the `src/load.php` resolution is exactly right, and the diff is back to the four files it should be.

I checked out `5c9abc8` and ran everything against the current `v3`, which now includes #872, #873 and #874:

| Check | Result |
| --- | --- |
| `composer cs` | clean |
| `composer phpstan` | clean, 74 files |
| `composer rector` | clean |
| Unit suite | 285 tests, 751 assertions, all passing |
| `PreReleaseNoticeTest` | 26 tests, 40 assertions |

## About those 38 deprecation errors

They do not reproduce here. I am on PHP 8.5.10 and the full suite is green, including the tests that use `setAccessible()`.

That looks like a stale `vendor/`: Composer resolves against whichever PHP it runs on, and `composer.lock` is gitignored, so installing on a different PHP than the one you then run the suite with can produce exactly this kind of mismatch. `AGENTS.md` describes the trap under the integration test section. A `rm -rf vendor && composer install` on the PHP you run the tests with should clear it. Nothing for you to fix in this PR either way.

## Review

I looked hardest at the parts most likely to be subtly wrong in a notice like this, and they are all correct:

- **`always_init()` is properly wired.** `src/load.php:116` calls it before the AJAX guard, so the dismiss handlers are registered even on requests where `init()` is skipped. The reasoning in your docblock holds.
- **The nonce channel lines up.** The script posts `_ajax_nonce`, and `check_ajax_referer()` without a `$query_arg` reads exactly that.
- **`is_pre_release()` behaves.** I ran it over our actual tag history and the edges: `3.0.0-beta.3`, `3.0.0-alpha.15` and `3.0.0-RC.1` come out as pre-releases; `3.0.0`, `2.11.12`, `3.0.0+build` and `3.0.0-` as stable; and `3.0.0-beta.2+build` as a pre-release, which is what the docblock promises.
- **The no-JS path works.** The Dismiss link is a real `admin-post.php` URL with a nonce, and the script only calls `preventDefault()` on its own link, never on core's `.notice-dismiss`. Delegating from the container means core's dynamically inserted button still reaches the handler.
- **Hook suffix, capability check and packaging** are all right, and storing the dismissal per version means the next beta shows the notice again — which is what we want.

The script also matches the style of `assets/js/admin-tabs.js`, so nothing stands out there.

## Three small things, none blocking

1. The `.fail()` fallback navigates to the dismiss URL, which carries the nonce that was rendered with the page. If the AJAX call failed *because* that nonce had expired, the fallback hits the same wall and the user gets `wp_die()`. Rare, and the alternative is more machinery than it is worth — worth a follow-up at most.
2. `notice.style.display = 'none'` hides the element rather than removing it. Core removes its own; a small inconsistency in the DOM, no user-visible difference.
3. On a network admin `plugins.php`, `manage_options` is not quite the right capability — `manage_network_options` is. It only means a network admin may see the notice on a page where we did not plan for it, which for a notice is harmless.

## Next step

No CI has run on this yet — there are zero check runs on `5c9abc8`. Because the branch is on a fork, the workflows need approving before they start; that is what the "blocked" state on the PR is about, not a conflict. I will approve the run.

One scheduling note: this is on the `3.0.0-beta.3` milestone, and that release is going out today with what is already merged. If the run comes back green in time it goes in; otherwise it moves to `3.0.0-beta.4`, which is days away, not weeks. Either way the work is done and nothing here needs changing.

2ndkauboy:
Correcting one of my three small points above: the capability one was wrong, and in your favour.

I wrote that a network admin might see the notice on network `plugins.php`, where `manage_options` is not the right capability. They cannot see it at all. Core dispatches the notice hooks as an `if / elseif / else` (`wp-admin/admin-header.php:293-313`, checked against 7.1):

```php
if ( is_network_admin() ) { do_action( 'network_admin_notices' ); }
elseif ( is_user_admin() ) { do_action( 'user_admin_notices' ); }
else { do_action( 'admin_notices' ); }
```

So in network admin `admin_notices` never fires, and nothing this PR registers can render there. The capability can never be consulted in that context, which makes the point moot rather than minor.

What it actually describes is the multisite gap already tracked in #866 — and the two are one piece of work: hooking `network_admin_notices` is the change that would make the capability check matter, since `manage_options` is the wrong gate there. I have noted that on #866 so it does not have to be rediscovered.

Nothing to change here. Also: CI has finished and all 18 checks pass.

PR #852, opened 2026-08-24
fix: add honeypot support for an input comment field

### LINKED ISSUE #738: Honeypot: support an `<input>` field as an alternative to `<textarea>`
Extracted from a code TODO at `src/Helpers/Honeypot.php:135`.

The honeypot currently renders a hidden `<textarea>`. Add the option to use an `<input>` element instead, for themes/setups where a textarea is problematic.

faisalahammad:
## Summary
Add honeypot support for themes that render the comment field as a text `<input>` instead of a `<textarea>`. Previously the honeypot renderer only handled a textarea, so an input-based comment form got no honeypot at all. The plugin now injects a honeypot that matches the field type automatically, with no new setting.

Fixes #738

### src/Helpers/Honeypot.php
Add an `input` case to the `switch` in `inject()`.

For a textarea form nothing changes. For an input form, the visible input keeps its id and class but gets an obfuscated name, and a hidden duplicate input that carries `name="comment"` is appended right after it as the bait. That mirrors how the textarea case traps the honeypot, so the existing check side of the rule keeps working unchanged.

**Why:** themes that output the comment field as `<input type="text">` had no honeypot, so honeypot-dependent spam protection was silently missing on those sites.

## Testing
**Test 1: Textarea form regression**
1. Leave the default `<textarea name="comment" id="comment">` in place.
2. View source, confirm a hidden honeypot textarea is present.
3. Blank submit: normal empty error, no spam flag.
4. Normal comment: posts and is not flagged as honeypot spam.
5. Fill the hidden honeypot textarea: comment flagged spam.
Result: unchanged behavior from before this fix.

**Test 2: Input form**
1. Replace the comment field with `<input type="text" id="comment" name="comment" class="comment-field">`.
2. View source, confirm the visible input keeps id and class but has an obfuscated name, and a hidden honeypot input with `name="comment"` is inserted after it.
3. Blank submit: normal empty error, no spam flag.
4. Normal comment: accepted, not flagged as honeypot spam.
5. Fill the hidden honeypot input: comment flagged spam.
Result: input-based comment forms now have a working honeypot.

**Test 3: Prefixed fields**
1. Add `<input id="comment-extra" name="comment-extra">` before the real comment field.
2. Confirm the real comment field is rewritten and `comment-extra` is left untouched.
Result: prefixed fields are not mistaken for the comment field.

## Screenshot
SCR-20260824-uhqr


2ndkauboy:
Thanks for this — a heads-up that `v3` has moved underneath this branch and it now conflicts.

Two PRs landed that rewrite `src/Helpers/Honeypot.php` substantially:

- **#883** added a marker field. `inject()` now emits a hidden `get_marker_name_for_post()` input on the path where it actually placed the secret field, and `Rules\Honeypot::precheck()` treats a missing secret field as an invalid request *only* when that marker is present. Before, any form the injection had skipped produced a final spam verdict, which misclassified every legitimate comment on themes the honeypot could not rewrite.
- **#834** stopped deriving the field names from the salt. `get_salt()` is gone; the secret is stored once in the `antispam_bee_honeypot_secret` option, and `get_secret_name_for_post()`, `get_secret_id_for_post()` and `get_marker_name_for_post()` all build on `get_secret()`.

**One thing worth knowing before you rebase**, because it is easy to miss: the new `case 'input':` branch needs to emit the marker too. Without it, an input-based comment form counts as "the honeypot was never injected", so `precheck()` casts no verdict and the honeypot is silently not enforced on exactly the themes this PR is meant to support. The `textarea` branch shows the shape — a hidden `<input>` named `self::get_marker_name_for_post()` with value `1`, added only on the success path.

Expect conflicts in `tests/Unit/Helpers/HoneypotHelperTest.php` as well; both PRs added tests there.

Since "Allow edits by maintainers" is off for this branch we cannot rebase it for you, so this needs a push from your side. No rush — I have moved it to the `3.0.0-beta.4` milestone, since `3.0.0-beta.3` is due today and this cannot realistically make it. #883 also means the failure mode this PR addresses is no longer a false-positive risk in the meantime: unsupported forms now degrade to "no verdict" rather than being marked as spam.

faisalahammad:
Rebased onto latest v3 and resolved the Honeypot.php conflict. Added marker emission to the input branch as requested. I also removed the old stacked notice commits, so this PR now contains only the honeypot change.

PR #853, opened 2026-08-24
fix: disable inactive rule checkboxes in delete-for-reasons

### LINKED ISSUE #740: Settings UI: disable reason checkboxes when the related rule is inactive
Extracted from a code TODO at `src/PostProcessors/DeleteForReasons.php:94`.

In the "delete for reasons" post processor, the per-reason checkboxes should be disabled when the corresponding rule is not active, so users can't configure a reason that can never trigger.

faisalahammad:
## Summary
In the post-processor options, a reason whose underlying rule is inactive in the rules section used to render as an enabled checkbox. A user could tick it, the choice would be stored, but the rule never fires, so the stored reason was dead weight with no UI hint that it can never trigger.

`DeleteForReasons::get_options()` now builds a `disabled_keys` map per reaction type. Any rule that implements `Controllable` and is inactive for that type is marked disabled. `CheckboxGroup::render()` reads the optional `disabled_keys` key and emits a `disabled` input plus a dimmed label. Non-controllable rules (no `is_active` method) keep their existing behaviour. The `sanitize` closure still keys on the full option list, so a stored value for a disabled reason survives the round-trip and is restored when the rule is re-enabled.

Fixes #740

### `src/PostProcessors/DeleteForReasons.php`
Build `$disabled_keys` per reaction type from controllable rules whose `is_active( $reaction_type )` is falsy. The new key is added to the emitted option entry alongside the existing `options` map. `sanitize` is unchanged on purpose.

**Why:** makes the post-processor options section a faithful view of what can actually happen at runtime, without silently dropping stored preferences.

### `src/Admin/Fields/CheckboxGroup.php`
Reads `$this->option['disabled_keys'] ?? []` before the row loop. When a row's slug is in the map, the input gets a `disabled` attribute and the wrapping label gets a new `asb-checkbox-group-disabled` class.

**Why:** opt-in renderer change. No other call site sets `disabled_keys`, so CountrySpam and LangSpam output is byte-equivalent to today.

### `assets/css/admin.css`
One new rule: `.asb-checkbox-group-disabled { color: #8c8f94; }`. Reuses the existing secondary-text colour used elsewhere in the admin stylesheet.

### Tests
- `tests/Unit/PostProcessors/DeleteForReasonsTest.php` (new, 6 cases): a controllable rule is disabled only when inactive, an active controllable rule is never disabled, a non-controllable rule is never disabled, an option list with no controllable rules yields an empty `disabled_keys`, the map is built per reaction type (not collapsed across the form), and `sanitize` still accepts a posted value for a disabled key.
- `tests/Unit/Admin/Fields/CheckboxGroupTest.php` (new, 5 cases): output is unchanged when `disabled_keys` is absent (regression guard for other call sites), a disabled row renders the `disabled` attribute plus the label class and only that row, a stored `on` value still renders as `checked="checked"` while the row is disabled, a stored `off` value renders unchecked but still disabled, and the row's `name` / `id` are the full controllable path with the slug suffix and are escaped.

## Testing

**Test 1: Manual round-trip on a fresh WordPress site**
1. Install `antispam-bee.740-test.zip` (built from this branch) via `Plugins → Add New → Upload Plugin`, then activate.
2. Open `Settings → Antispam Bee`, expand the **Rules** section, confirm "BBCode link" is active.
3. Open the **"Delete by reasons"** section. Tick the BBCode reason. Save.
4. Re-open the same section. The BBCode row is `checked="checked"` and enabled (pre-existing baseline).
5. Go back to the **Rules** section and deactivate "BBCode link". Save.
6. Re-open **"Delete by reasons"**. Expected: BBCode row is still listed and still `checked="checked"`, but the checkbox is `disabled` and the label text is dimmed. Clicking or unchecking the row has no effect.
7. Re-activate "BBCode link" in the **Rules** section. Re-open **"Delete by reasons"**. Expected: the BBCode row is enabled again, the stored `on` value is honoured, no further user action is needed.

**Test 2: Per-reaction-type check**
1. In the **Rules** section, deactivate "BBCode link" for the **Comment** reaction type only (per-type sub-checkboxes). Save.
2. Open the **"Delete by reasons"** section. The BBCode row is disabled under "Comment" and enabled under "Linkback" (if the site has linkbacks).

This proves the `disabled_keys` map is built per reaction type, not collapsed across the form.

**Test 3: Automated checks in the worktree**
- `composer test:unit` 199 tests, 0 failures, 0 regressions (the 11 errors in the output are pre-existing PHP 8.5 `ReflectionProperty::setAccessible()` deprecations in `tests/Unit/Helpers/DebugModeTest.php`, unrelated to this change).
- `composer lint-php` clean.
- `composer phpstan` level 8 clean.

## Edge cases covered by unit tests
- A rule that does not implement `Controllable` (e.g. `asb-empty`, `asb-deny-country`) is never disabled. `is_active()` does not exist on it, and `ComponentsHelper::filter()` treats it as always-on.
- A rule whose stored value is `'off'` (the user explicitly unticked it before deactivation) renders unchecked but still disabled.
- An option list with zero controllable rules yields an empty `disabled_keys` array. The disable branch is correctly short-circuited.
- Sanitization still accepts a posted value for a disabled key, so a direct form POST that re-sends the slug does not silently lose the stored reason.

## Notes for reviewer
- Schema change is additive. `disabled_keys` defaults to `[]` in the renderer, no migration needed.
- Stored data is preserved. Re-enabling a rule restores the user's prior selection without re-ticking.
- No new i18n strings. The `disabled` attribute and the `asb-checkbox-group-disabled` class are not user-facing copy.
- Backward compatible: no breaking changes to public PHP, JS, or stored option shape.

## Screen recording
https://videos.faisalahammad.com/recordings/WblPUVF5Okuc2YYIBfFL

https://github.com/user-attachments/assets/e8b8beae-55cd-4eb7-b63d-154c8bda94a6


2ndkauboy:
Thanks for this one too. It conflicts with `v3` now, but the conflicts are inherited rather than caused by the change itself.

This branch is stacked: it carries the commits of #850 and #852 as well as its own `fix: disable inactive rule checkboxes in delete-for-reasons`. All of the conflicts come from #852's honeypot commit, because #883 and #834 rewrote `src/Helpers/Honeypot.php` in the meantime — `get_salt()` is gone, the secret now lives in the `antispam_bee_honeypot_secret` option, and `inject()` emits a marker field that `Rules\Honeypot::precheck()` requires before it treats a missing secret field as an invalid request. The details are in the comment on #852, including a note about the new `case 'input':` branch needing to emit that marker.

So the order matters: rebase #852 onto `v3` first, then rebase this branch on top of it. Rebasing this one on its own would mean resolving the same honeypot conflicts twice and the two PRs would drift apart.

Your own change here does not touch the honeypot logic, so once the inherited commits are sorted out this should apply cleanly.

"Allow edits by maintainers" is off, so we cannot do the rebase for you. I have moved this to the `3.0.0-beta.4` milestone — `3.0.0-beta.3` is due today and a stacked fork PR cannot realistically land in time.

faisalahammad:
Rebased on top of the rebased #852 as requested. The change for inactive reason checkboxes replayed cleanly.

PR #862, opened 2026-08-26
fix: stop overwriting a supplied comment IP and drop the inert invalid-request key

2ndkauboy:
Two leftovers in `Comment::process()`.

`$reaction['ab_spam__invalid_request']` was written but never read. The rule that carries this name, `Rules\InvalidRequest::verify()`, reads `$_POST['ab_spam__invalid_request']`, and `build_payload()` does not copy the key into the normalized payload, so nothing consumed it. It is removed rather than wired up to the channel the rule reads: "SCRIPT_NAME did not parse, therefore definitive spam" is not a heuristic worth acting on now that the handler deliberately serves entry points other than the comment form, where SCRIPT_NAME semantics vary by SAPI. The rule keeps working through `Honeypot::precheck()`, which sets that key for real comment-form submissions missing the secret field — the meaningful signal.

`comment_author_IP` was replaced unconditionally, for every comment-type reaction, before the handler had decided whether it processes the reaction at all. Because the mutated array is what goes back to `preprocess_comment`, the overwrite applied even to comments the plugin never classifies. Core only falls back to `REMOTE_ADDR` for an absent value, so importers, migrations and plugins passing an explicit or historical IP had it silently replaced with the IP of whoever ran the request — and ASB hooks at priority 1, so it won against later filters. `IpHelper::get_client_ip()` also returns '' for anything `FILTER_VALIDATE_IP` rejects, such as a zone-scoped IPv6 address, so a value core would have stored could be blanked, degrading both core's disallowed-list IP matching and the `DbSpam` rule.

The assignment now happens after the skip check, fills only an empty value, and never writes an empty string over an existing one.

### DIFF
--- a/src/Handlers/Comment.php
+++ b/src/Handlers/Comment.php
@@ -60,17 +60,25 @@ class Comment extends Reaction {
 			return $reaction;
 		}
 
-		$reaction['comment_author_IP'] = IpHelper::get_client_ip();
-
-		$request_uri  = isset( $_SERVER['SCRIPT_NAME'] ) ? esc_url_raw( wp_unslash( $_SERVER['SCRIPT_NAME'] ) ) : '';
-		$request_path = DataHelper::parse_url( $request_uri, 'path' );
-
-		if ( empty( $request_path ) ) {
-			$reaction['ab_spam__invalid_request'] = 1;
-		}
-
 		if ( self::skip_verification( $reaction ) ) {
 			return $reaction;
+		}
+
+		/*
+		 * Only filled in when the caller did not supply one. Core does the same
+		 * (`wp_new_comment()` falls back to REMOTE_ADDR only for an absent value), so
+		 * importers, migrations and plugins that pass a historical or explicit IP keep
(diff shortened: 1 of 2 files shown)


2ndkauboy:
Stacked on #858 — based on `fix/comment-rules-bypassed-outside-comments-post` rather than `v3`, because the changes sit inside the `skip_verification()` guard that #858 introduces. Retarget to `v3` once #858 has been merged.

2ndkauboy:
Superseded by #884, which is this branch rebased onto `v3`.

To record what happened: this pull request was not closed on purpose. It was stacked on #858, and deleting that branch when #858 merged fired `base_ref_deleted`, which closed this one two seconds later — the timeline shows both events at `2026-09-20T13:56:13Z`. The fix never reached `v3`; `src/Handlers/Comment.php:63` still carried the unconditional `comment_author_IP` overwrite until now.

GitHub refuses to reopen a pull request whose base branch no longer exists, so the work continues in #884 rather than here. Same branch, same commit, rebased with `--onto v3` to drop #858's commits (they landed squashed as a92e370).

